PR #27, opened 2015-01-14
Stack & globals instrumentation

aryabinin:


arch/x86/include/asm/kasan.h
```diff
@@ -1,8 +1,9 @@
 #ifndef _ASM_X86_KASAN_H
 #define _ASM_X86_KASAN_H
 
-#define KASAN_SHADOW_START	0xffffd90000000000UL
-#define KASAN_SHADOW_END	0xffffe90000000000UL
+#define KASAN_SHADOW_START      (KASAN_SHADOW_OFFSET + (1ULL << 61) \
```
dvyukov: This puzzled me for a while.
Please add a comment explaining that kernel addresses do not start at zero, and so we add kernel_address_space_start>>kasan_shadow_scale to the offset that compiler uses.

arch/x86/kernel/head_64.S
```diff
@@ -528,11 +528,11 @@ NEXT_PAGE(zero_pud)
 	FILL(zero_pmd - __START_KERNEL_map + __PAGE_KERNEL_RO, 512)
 
 NEXT_PAGE(poisoned_pte)
-	FILL(poisoned_page - __START_KERNEL_map + __PAGE_KERNEL_RO, 512)
+	FILL(poisoned_page - __START_KERNEL_map + _KERNPG_TABLE, 512)
```
dvyukov: The page is not poisoned anymore. Please rename s/poisoned/kasan_early/ or something along these lines.

arch/x86/kernel/head_64.S
```diff
@@ -528,11 +528,11 @@ NEXT_PAGE(zero_pud)
 	FILL(zero_pmd - __START_KERNEL_map + __PAGE_KERNEL_RO, 512)
```
dvyukov: I would also be nice to prefix the zero page with kasan_ as it is used only in kasan, which is a quite special execution mode. So that kernel developers are not puzzled about its purpose.

arch/x86/mm/kasan_init_64.c
```diff
@@ -104,5 +194,33 @@ void __init kasan_init(void)
...
+
+		if (!IS_ENABLED(CONFIG_KASAN_GLOBALS))
+			populate_zero_shadow(kasan_mem_to_shadow(MODULES_VADDR),
+					KASAN_SHADOW_END);
+		else
+			populate_zero_shadow(kasan_mem_to_shadow(MODULES_END),
+					KASAN_SHADOW_END);
+	}
+
+	memset(poisoned_page, 0xF9, PAGE_SIZE);
```
dvyukov: s/0xF9/KASAN_SHADOW_GAP/

dvyukov: Please either use another page as poisoned page, or comment on its declaration that it is used as 'scratch' shadow during early bootstrap when kasan is disabled, and later reused as poisoned page. I am not sure how it should be named in the latter case.

kernel/module.c
```diff
@@ -1827,6 +1828,7 @@ static void unset_module_init_ro_nx(struct module *mod) { }
 void __weak module_free(struct module *mod, void *module_region)
 {
 	vfree(module_region);
+	kasan_module_free(module_region);
```
dvyukov: This is a wrong place to insert the call. module_free is __weak and intended as default implementation unless arch wants to override it. A bunch of arch in fact override it.

mm/kasan/kasan.c
```diff
@@ -245,6 +247,26 @@ static __always_inline void check_memory_region(unsigned long addr,
 	kasan_report(addr, size, write);
 }
 
+int kasan_module_alloc(void *addr, size_t size)
+{
+
+	size_t shadow_size = round_up(size >> KASAN_SHADOW_SCALE_SHIFT,
+				PAGE_SIZE);
+	unsigned long shadow_start = kasan_mem_to_shadow((unsigned long)addr);
```
dvyukov: please assert that shadow_start is page aligned
this is important here

mm/kasan/kasan.h
```diff
@@ -20,6 +36,25 @@ struct access_info {
 	unsigned long ip;
 };
 
+struct asan_source_location {
```
dvyukov: We don't yet have names prefixed with 'asan_' in kernel. I think it's better to rename it to 'kasan_' to keep it consistent with the rest of the names here, even if it diverges from user-space naming.

dvyukov:
> If we add a lot of config options, it will be a hell to support all possible configurations.
> It doesn't look too bad IMO.

Please-please-please remove the configs.
Configs are evil all others equal. We already have 8 different configurations. Have somebody tested them at least once?
Configs should be added when there is a strong need at hand and no ways around. I do not mind to add them in that situation, but not now.
We will get complaints from users about broken why-in-this-world-one-would-ever-use configurations.
We will get complaints about missed bugs when a user actually disabled detection long time ago without any particular reason and without understanding what he is doing, it just looked like a "good" idea for him at that time.
We should strongly push for a one-configuration-fits-all. And diverge from that only when we understand the actual need (and setup testing for that additional configuration). Rather than adding all possible might-be-useful-in-future configurations proactively.

arch/x86/mm/kasan_init_64.c
aryabinin: Another page will require additional 3 pages for PTEs, PMDs, PUDs, so I tend to adding some comments.
I think poisoned_page is fine (or kasan_poisoned_page even better) as it used as poisoned page most of the kernel's lifetime.

kernel/module.c
aryabinin: Sorry, but I didn't get this. This seems like a right place to me for arches that use default module_free() (x86).

 How from 
> module_free is __weak and intended as default implementation unless arch wants to override it

follows that

> This is a wrong place to insert the call.

Please elaborate.

aryabinin:
> Please-please-please remove the configs.
Configs are evil all others equal. We already have 8 different configurations. Have somebody tested them at least once?
Configs should be added when there is a strong need at hand and no ways around. I do not mind to add them in that situation, but not now.
We will get complaints from users about broken why-in-this-world-one-would-ever-use configurations.
We will get complaints about missed bugs when a user actually disabled detection long time ago without any particular reason and without understanding what he is doing, it just looked like a "good" idea for him at that time.
We should strongly push for a one-configuration-fits-all. And diverge from that only when we understand the actual need (and setup testing for that additional configuration). Rather than adding all possible might-be-useful-in-future configurations proactively.

That sounds convincing. I'm certainly don't mind of dropping KASAN_GLOBALS config.
But I'm not sure how to handle KASAN_STACK, since stack instrumentation won't work with gcc 4.9.2.

Supporting 4.9.2 becoming a bit painful now. I'm not sure that we have a right to drop 4.9.2 support, but this would make our life easier.

aryabinin:
Updated.

Two pending issues remain: 
* what's wrong with module_free()?
* what we gonna do with CONFIG_KASAN_STACK?

dvyukov:
> what's wrong with module_free()?

I think I see the point of misunderstanding.
I treated kasan_module_load/unload callbacks as platform-independent, and put them into module.c code itself so that all arches do not need to care about it.
You made them arch-specific and put into arch-specific code. From this point of view your code is correct.

Is there any reason to move this to arch-specific code? Otherwise keeping it in common code looks better to me.

dvyukov:
> what we gonna do with CONFIG_KASAN_STACK?

How exactly it fails with gcc4.9?
I thought that Makefile takes care of removing the unsupported flag, and the compiler just do not emit instrumentation. Which is fine, we just don't catch bugs there.
We also map shadow for stack, which can be avoided. But it does not look like a big deal to me. We want people to move to 5.0 because it also has inline instrumentation.

aryabinin:
Yep, you are right, it shouldn't fail.
My concern is - does it worth to keep this option, so 4.9.2 users (do they even exist?) could disable it and save some stack?
But seems we don't care much about them, and removing option is preferable, right?

dvyukov:
> But seems we don't care much about them, and removing option is preferable, right?

Yes, it is preferable.
There are chances that (1) we will need more modifications to gcc and they won't be in 4.9 or (2) by the time we upstream and polish kasan, gcc 5.0 will be released by distros.

aryabinin:
Removed stack config

PR #43, opened 2020-04-03
kfence: initial allocation hookup prototype

ramosian-glider:
No error checking for now, we just steal the freelist from a random
kmalloc cache from time to time, and route SLUB's slow path into our
allocator that returns pages.


ramosian-glider:
@melver please take a look.
I took the code from my github and changed `guarded_alloc()` and `guarded_free()` to work with a fixed page pool.

PR #45, opened 2020-04-03
kfence: Various cleanups

melver:
* Add missing 'inline' in header.
* Add missing declarations in header.
* Add some TODOs.
* clang-format.

Signed-off-by: Marco Elver <elver@google.com>

### DIFF
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -1,26 +1,36 @@
 // KFENCE api
 
+#include <linux/types.h>
+
+struct kmem_cache;
+struct page;
+
 #ifdef CONFIG_KFENCE
+/* TODO: API documentation */
+
 void kfence_init(void);
+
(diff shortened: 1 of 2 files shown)


ramosian-glider:
LGTM, thank you!
By the way, how do you run clang-format on the kernel source?

melver:
> LGTM, thank you!
> By the way, how do you run clang-format on the kernel source?

The kernel already has a .clang-format file, so it should just be a matter of running clang-format -i <file>. I'm also using an editor plugin (vim), so I can run it on the buffer I'm editing as well as select individual lines to format. I found that clang-format isn't necessarily the source-of-truth for formatting, but works as a good first approximation while some piece of code is in development; later I go and change some things (e.g. remote 80-char restriction when it makes things more readable as was suggested by maintainers for initial KCSAN PR).

PR #46, opened 2020-04-04
test_kfence.c + more slots to store stolen freelists

ramosian-glider:


lib/test_kfence.c
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Test cases for SL[AOU]B/page initialization at alloc/free time.
```
melver, line 3: I'm currently working on KCSAN tests using KUnit+Torture frameworks. We probably need to change this to use those eventually as well. But for now this is probably fine.

I'm happy to convert once I figured out the KCSAN tests.

mm/kfence.c
```diff
@@ -152,28 +158,44 @@ void guarded_free(void *addr)
...
-			page = virt_to_page(ret);
-		}
-		spin_unlock_irqrestore(&kfence_lock, flags);
-		pr_info("kfence_alloc_and_fix_freelist returns %px\n", ret);
-		return ret;
-	}
+	void *freelist;
+
+	fl = find_freelist(s);
+	if (fl == -1)
```
melver, line 237: It'd be helpful if you could add some rudimentary comments for non-trivial things.

AFAIK, you're doing double-checked locking here, right?

mm/kfence.c
```diff
@@ -152,28 +158,44 @@ void guarded_free(void *addr)
 	spin_unlock_irqrestore(&kfence_alloc_lock, flags);
 }
 
+static int find_freelist(struct kmem_cache *c)
```
melver, line 218: This should return a pointer, and just NULL if none found.

lib/test_kfence.c
ramosian-glider: I just wrote a quick-and-dirty test to debug error reporting, but I agree KUnit is a must.

ramosian-glider: (also the above snippet contains an irrelevant comment, I'll fix that)

mm/kfence.c
```diff
@@ -103,6 +107,20 @@ static void __meminit kfence_protect(unsigned long addr)
...
 
+static void __meminit kfence_unprotect(unsigned long addr)
+{
+	unsigned long addr_end;
+	pte_t *pte;
+	unsigned int level;
+
+	addr_end = addr + PAGE_SIZE;
+	pte = lookup_address(addr, &level);
+	BUG_ON(!pte);
```
melver, line 118: We'll probably have to get rid of the BUG_ONs eventually. checkpatch will remind us.

mm/kfence.c
```diff
@@ -18,10 +18,14 @@ struct stored_freelist {
 	struct kmem_cache *cache;
 	void *freelist;
 };
+
+enum kfence_object_state { KOS_UNUSED, KOS_ALLOCATED, KOS_FREED };
```
melver, line 22: KOS is a bit cryptic. Maybe just 'KFENCE_OBJECT_' ?

mm/kfence.c
ramosian-glider: Right. Currently the locking is somewhat messy, I'll need to revisit it.
Guess for now double-checked locking is actually an overkill, since we haven't started measuring performance.

mm/kfence.c
ramosian-glider: Agreed. I thought it'll be used very often, but it doesn't seem to be the case.

mm/kfence.c
ramosian-glider: We probably need to bail out early on syzbot and during testing, but keep the kernel working in production.
I'll replace these checks with a custom assertion that we can later make configureable.

melver: A if (WARN_ON_ONCE(..)) and some error-recovery path should be enough. On syzbot with panic_on_warn that'll be what we want.

mm/kfence.c
ramosian-glider: Good idea, done.

mm/kfence.c
ramosian-glider: Dropped the double-checked locking.

PR #48, opened 2020-04-08
Merge recent fixes from kfence-glider

ramosian-glider:


mm/kfence.c
```diff
@@ -79,6 +81,21 @@ static inline depot_stack_handle_t save_stack(gfp_t flags)
...
+{
+	pr_err("Disabling KFENCE\n");
+	WRITE_ONCE(kfence_enabled, false);
+}
+
+#define kfence_assert(cond)                                                    \
+	do {                                                                   \
+		if (unlikely(!(cond))) {                                       \
+			WARN(1, "KFENCE assertion failure at %s:%d\n",         \
+			     __FILE__, __LINE__);                              \
```
melver, line 94: WARN already has file and line number.

Why not just 'if (WARN(!(cond), "KFENCE assertion failure")' ?

melver: A simpler solution would be to just have an expression macro:

#define KFENCE_WARN_ON(cond) \({ \
   bool __cond = WARN_ON(cond); \
   if (__cond)\ 
     kfence_disable(); \
  __cond; \
})

Because every assertion failure or warning should go into an error-recovery path.

mm/kfence.c
```diff
@@ -95,7 +112,7 @@ static void __meminit kfence_protect(unsigned long addr)
 
 	addr_end = addr + PAGE_SIZE;
 	pte = lookup_address(addr, &level);
-	BUG_ON(!pte);
+	kfence_assert(pte);
```
melver, line 115: This needs some error-recovery. I assume that this code cannot just continue, otherwise it'll try to deref pte below.

mm/kfence.c
```diff
@@ -119,8 +136,8 @@ static void __meminit kfence_unprotect(unsigned long addr)
 
 	addr_end = addr + PAGE_SIZE;
 	pte = lookup_address(addr, &level);
-	BUG_ON(!pte);
-	BUG_ON(level != PG_LEVEL_4K);
+	kfence_assert(pte);
```
melver, line 139: There should be some error recovery.

mm/kfence.c
```diff
@@ -363,6 +380,12 @@ bool kfence_handle_page_fault(unsigned long addr)
 	if ((addr < kfence_pool_start) || (addr >= kfence_pool_end))
 		return false;
 
+	if (!READ_ONCE(kfence_enabled)) {
```
melver, line 388: Why does this need to be here? Should this be an error-recovery path for an assertion/warn failure?

mm/kfence.c
melver: With the suggestion for KFENCE_WARN, it'd be 'if (KFENCE_WARN(!pte)) { bail out; }'

mm/kfence.c
ramosian-glider: Yes, you are right. I actually just blindly replaced BUG_ONs with the new assertion without thinking whether the recovery is needed. I'll fix that.

mm/kfence.c
ramosian-glider: Yes, I'll make the function return bool to track the error status.

mm/kfence.c
ramosian-glider: Done.

mm/kfence.c
ramosian-glider: If the tool is disabled, we don't want to report any new errors.
However there may be leftover protections causing page faults. If we want the kernel to proceed we need to make those pages accessible.

melver:
LGTM, thanks!

PR #49, opened 2020-04-10
kfence: get rid of per-cpu variables

ramosian-glider:
Those didn't make sense in the first place: if we stole a freelist on
CPU0 and switched to CPU1 before reinstating it, the freelist contents
would've been stuck on CPU0.

### DIFF
--- a/lib/test_kfence.c
+++ b/lib/test_kfence.c
@@ -45,6 +45,8 @@ static int do_test_oob(size_t size)
 	char *c;
 
 	buffer = alloc_from_kfence(size, GFP_KERNEL);
+	if (!buffer)
+		return 1;
 	c = ((char *)buffer) + size + 1;
 	READ_ONCE(*c);
 	kfree(buffer);
@@ -57,6 +59,8 @@ static int do_test_uaf(size_t size)
 	char *c;
 
 	buffer = alloc_from_kfence(size, GFP_KERNEL);
+	if (!buffer)
+		return 1;
 	c = (char *)buffer;
 	kfree(buffer);
 	READ_ONCE(*c);
--- a/lib/test_kfence.c
+++ b/lib/test_kfence.c
@@ -6,47 +6,60 @@
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
 
 #include <linux/delay.h>
+#include <linux/mm.h>
 #include <linux/module.h>
 #include <linux/slab.h>
+#include <linux/slab_def.h>
+
+static bool is_kfence_allocation(void *ptr)
+{
+	struct page *page = virt_to_page(ptr);
+
+	if (!page || !PageSlab(page))
+		return false;
+	if (page->slab_cache && page->slab_cache->name)
+		return !strcmp(page->slab_cache->name, "kfence_slab_cache");
+	return false;
+}
 
 #define MAX_ITER 100
-int do_test_oob(size_t size)
+static void *alloc_from_kfence(size_t size, gfp_t gfp)
 {
+	void *res;
 	int i;
-	volatile char *c;
-	void **buffers;
 
-	buffers = kmalloc_array(MAX_ITER, sizeof(void *), GFP_KERNEL);
 	for (i = 0; i < MAX_ITER; i++) {
-		buffers[i] = kmalloc(size, GFP_KERNEL);
-		c = ((char *)buffers[i]) + size + 1;
-		(void)*c;
+		res = kmalloc(size, gfp);
+		if (is_kfence_allocation(res))
+			return res;
+		kfree(res);
 		/* TODO: sleep time depends on heartbeat period. */
 		msleep(100);
 	}
-	for (i = 0; i < MAX_ITER; i++) {
-		kfree(buffers[i]);
-	}
-	kfree(buffers);
+	return NULL;
+}
+
+static int do_test_oob(size_t size)
+{
+	void *buffer;
+	char *c;
+
+	buffer = alloc_from_kfence(size, GFP_KERNEL);
+	c = ((char *)buffer) + size + 1;
+	READ_ONCE(*c);
+	kfree(buffer);
 	return 0;
 }
 
-int do_test_uaf(size_t size)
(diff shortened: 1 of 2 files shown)


ramosian-glider:
Added use-after-free detection.

mm/kfence.c
```diff
@@ -85,11 +85,12 @@ static inline void kfence_disable(void)
...
 
-#define KFENCE_WARN_ON(cond) ({	\
-	bool __cond = WARN_ON(cond);	\
-	if (__cond) \
-	kfence_disable(); \
-	__cond; \
+#define KFENCE_WARN_ON(cond)                                                   \
+	({                                                                     \
+		bool __cond = WARN_ON(cond);                                   \
+		if (__cond)                                                    \
```
melver, line 104: No idea of the compiler is clever enough to have the 'unlikely' in WARN_ON propagates here. However, if this is only used in slow-path, no action required.

lib/test_kfence.c
```diff
@@ -31,11 +31,31 @@ int do_test(size_t size)
...
 
+int do_test_uaf(size_t size)
+{
+	int i;
+	volatile char *c;
+	void **buffers;
+
+	buffers = kmalloc_array(MAX_ITER, sizeof(void *), GFP_KERNEL);
+	for (i = 0; i < MAX_ITER; i++) {
+		buffers[i] = kmalloc(size, GFP_KERNEL);
```
melver, line 42: I don't understand why you need 'buffers[i]'. They do not appear to be reused in later iterations or elsewhere.

Why not just 'char *buffer = kmalloc...' ?

lib/test_kfence.c
```diff
@@ -31,11 +31,31 @@ int do_test(size_t size)
...
+		kfree(buffers[i]);
+		c = ((char *)buffers[i]);
+		(void)*c;
```
melver, line 45: You could just make this 'READ_ONCE(*buffer)'.

melver: Or, if you want to search the whole buffer until the first random "x" appears, something like 'strnstr(buffer, "x", size)'.

lib/test_kfence.c
```diff
@@ -31,11 +31,31 @@ int do_test(size_t size)
 	return 0;
 }
 
+int do_test_uaf(size_t size)
```
melver, line 34: Should these be static?

Not that it matters, since we'll rewrite this for kunit anyway (I'm happy to rewrite to kunit, so no action required by you).

mm/kfence.c
```diff
@@ -57,6 +57,19 @@ struct kfence_freelist_t {
 	struct list_head list;
 	void *obj;
 };
+/*
+ * kfence_freelist_t is a wrapper around kfence page pointers that allows
+ * chaining them.
+ * |kfence_freelist| is a FIFO queue of non-allocated pages, |kfence_recycle| is
```
melver, line 63: The kernel uses '@' to refer to variables in documentation. E.g. "@kfence_freelist".

mm/kfence.c
```diff
@@ -254,18 +262,20 @@ void *guarded_alloc(size_t size, gfp_t gfp)
 		item = list_entry(kfence_freelist.list.next,
 				  struct kfence_freelist_t, list);
 		obj = item->obj;
-		kfence_freelist.list.next = item->list.next;
+		list_del(&(item->list));
```
melver, line 265: list_del followed by list_add is undefined. If you turn on CONFIG_DEBUG_LIST you will likely get warnings.

I think this needs to be list_del_init

mm/kfence.c
```diff
@@ -292,12 +303,14 @@ void guarded_free(void *addr)
 	item = list_entry(kfence_recycle.list.next, struct kfence_freelist_t,
 			  list);
 	item->obj = aligned_addr;
-	kfence_recycle.list.next = item->list.next;
-	list_add(&(item->list), &kfence_freelist.list);
+	list_del(&(item->list));
```
melver, line 306: Same here, should probably be list_del_init.

mm/kfence.c
ramosian-glider: I think we can just add the 'unlikely' to be on the safe side.

mm/kfence.c
ramosian-glider: I don't think this is necessary as long as the lists are locked. See e.g. https://marc.info/?l=linux1394-devel&m=114328842931957
CONFIG_DEBUG_LIST also didn't show any warnings.

mm/kfence.c
ramosian-glider: I thought it's for function arguments/struct members only. I don't have a strong preference though.
Fixed.

mm/kfence.c
ramosian-glider: ditto

lib/test_kfence.c
ramosian-glider: Yes, I rewrote this test a bit, and now buffers[i] aren't needed anymore.
Actually, I wanted to introduce a function that checks that a buffer belongs to kfence pool, so that we only test such allocations. Let me try that...

lib/test_kfence.c
ramosian-glider: I like READ_ONCE() more, thanks!
In the future we might want to precisely point at a byte in the buffer to ensure the error report is correct.

lib/test_kfence.c
ramosian-glider: Thanks!
I'll add static anyway.

ramosian-glider:
Fixed the comments.

lib/test_kfence.c
```diff
@@ -6,47 +6,60 @@
...
-	kfree(buffers);
+	return NULL;
+}
+
+static int do_test_oob(size_t size)
+{
+	void *buffer;
+	char *c;
+
+	buffer = alloc_from_kfence(size, GFP_KERNEL);
```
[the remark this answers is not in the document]
ramosian-glider: Done

PR #51, opened 2020-04-16
kfence: don't merge (for review)

dvyukov:


melver:
Does Github automatically figure out if both 'upstream' and 'to-pull' branch were updated? Because it looks like it doesn't and wants to merge *all* the commits from 5.6 to 5.8. :-)

ramosian-glider:
We could probably just drop this PR and open another one, there's no unique information associated with it.

melver:
Closing, because Github seems to get confused when both upstream and kfence are updated and seems to somehow make a mess of it.

PR #54, opened 2020-04-17
kfence: use get_random_int()

ramosian-glider:
Use get_random_int() to select a random kmalloc cache and to place the
allocated object at the beginning/end of the page.
Also fix ksize() to handle both cases of object placement within the
page.

### DIFF
--- a/mm/kfence.c
+++ b/mm/kfence.c
@@ -5,6 +5,7 @@
 
 #include <linux/mm.h> // required by slub_def.h, should be included there.
 #include <linux/moduleparam.h>
+#include <linux/random.h>
 #include <linux/slab.h>
 #include <linux/slub_def.h>
 #include <linux/spinlock_types.h>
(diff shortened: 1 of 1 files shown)


mm/kfence.c
```diff
@@ -271,6 +272,7 @@ void *guarded_alloc(size_t size, gfp_t gfp)
 	void *obj = NULL, *ret;
 	struct kfence_freelist_t *item;
 	int index = -1;
+	bool right = get_random_int() % 2;
```
melver, line 275: I think prandom_u32_max is simpler. I don't think we need the guarantees that get_random provides.

mm/kfence.c
```diff
@@ -506,23 +516,28 @@ bool kfence_handle_page_fault(unsigned long addr)
 	return kfence_unprotect(addr);
 }
 
+static struct kmem_cache *kfence_pick_cache(void)
+{
+	int index = get_random_int() %
```
melver, line 521: Just prandom_u32_max will be simpler.

mm/kfence.c
ramosian-glider: Fair enough. Done.

PR #56, opened 2020-04-21
Change OOB test, fix deadlocks

ramosian-glider:


mm/kfence.c
```diff
@@ -463,17 +464,27 @@ static void kfence_dump_object(int obj_index)
 	}
 }
 
+static inline void kfence_start_report(void)
+{
+	console_lock();
```
melver, line 469: Is this going to have other strange side-effects? Is this approved usage of 'console_lock()'?

At least a comment assuring that this is an ok thing to do would be good.

ramosian-glider: I think so. At least it is in the header, is exported (i.e. available to modules) and has an unambiguous comment: https://elixir.bootlin.com/linux/latest/source/kernel/printk/printk.c#L2278

I'll add a comment to our code to clarify that.

melver: What I'm worried about here is the "Acquires a lock which guarantees that the caller has exclusive access to the console system [...]", and then subsequent call to printk. What does "console system" here imply? I think it's the internal state of the console system, but says nothing about whether or not calling printk() still works.

The usage across the kernel I see is console drivers etc.

If we know that 'console_lock()' is always re-entrant, then we can probably make the claim that it's safe. If there is no guarantee that it's re-entrant, then we might eventually have a problem.

ramosian-glider: Turns out it is not re-entrant, at least calling console_lock() twice freezes the task :(

ramosian-glider:
Reverted the console_lock() hack, moved error reporting outside the critical section instead.
This could've been made cleaner by making the object index part of `struct alloc_metadata`, but we won't need it anywhere else except reporting.
Please take another look.

mm/kfence.c
```diff
@@ -448,48 +459,57 @@ static void kfence_print_stack(int obj_index, bool is_alloc)
...
-	kfence_dump_object(obj_index);
+	kfence_dump_object(obj_index, object);
 	pr_err("==================================================================\n");
 }
 
+static inline void copy_obj_metadata(int index, struct alloc_metadata *obj)
+{
+	if ((index < 0) || (index >= KFENCE_NUM_OBJ))
+		memset(obj, 0, sizeof(struct alloc_metadata));
+	memcpy(obj, &kfence_metadata[index], sizeof(struct alloc_metadata));
```
melver, line 505: *obj = kfence_metadata[index]
should be simpler and safer.

mm/kfence.c
```diff
@@ -448,48 +459,57 @@ static void kfence_print_stack(int obj_index, bool is_alloc)
...
 	dump_stack();
-	kfence_dump_object(obj_index);
+	kfence_dump_object(obj_index, object);
 	pr_err("==================================================================\n");
 }
 
+static inline void copy_obj_metadata(int index, struct alloc_metadata *obj)
+{
+	if ((index < 0) || (index >= KFENCE_NUM_OBJ))
+		memset(obj, 0, sizeof(struct alloc_metadata));
```
melver, line 504: If index is oob, it won't return here. Is this intentional?

mm/kfence.c
```diff
@@ -448,48 +459,57 @@ static void kfence_print_stack(int obj_index, bool is_alloc)
...
 	pr_err("==================================================================\n");
 	pr_err("BUG: KFENCE: use-after-free at address %px on object #%d\n",
 	       (void *)address, obj_index);
 	dump_stack();
-	kfence_dump_object(obj_index);
+	kfence_dump_object(obj_index, object);
 	pr_err("==================================================================\n");
 }
 
+static inline void copy_obj_metadata(int index, struct alloc_metadata *obj)
```
melver, line 501: Is this function even needed?

Is it ever going to be called with index <0 || index >= KFENCE_NUM_OBJ?

Why not just do
  object = kfence_metadata[index]
at the callsites, and remove this function?

mm/kfence.c
```diff
@@ -448,48 +459,57 @@ static void kfence_print_stack(int obj_index, bool is_alloc)
...
+	if ((index < 0) || (index >= KFENCE_NUM_OBJ))
+		memset(obj, 0, sizeof(struct alloc_metadata));
+	memcpy(obj, &kfence_metadata[index], sizeof(struct alloc_metadata));
+}
+
 bool kfence_handle_page_fault(unsigned long addr)
 {
 	int page_index, obj_index, report_index = -1, dist = 0, ndist;
 	unsigned long flags;
+	struct alloc_metadata object = {};
```
melver, line 512: AFAIK this 0-initializes object and the memset(obj, 0, ..) isn't needed, right?

mm/kfence.c
ramosian-glider: Ack.

mm/kfence.c
ramosian-glider: No. Good catch!

mm/kfence.c
ramosian-glider: I sometimes tend to over-generalize, just like here. Please stop me if you see that :)

mm/kfence.c
ramosian-glider: I dropped copy_obj_metadata() altogether.

ramosian-glider:
Dropped copy_obj_data(), will merge this PR now.
Thanks for the review!

PR #60, opened 2020-04-29
Cleanups

melver:


include/linux/kfence.h
```diff
@@ -1,4 +1,6 @@
-// KFENCE api
+/* SPDX-License-Identifier: GPL-2.0 */
+#ifndef _LINUX_KFENCE_H
```
ramosian-glider, line 2: Yes, completely missed this. Thank you!

include/linux/kfence.h
```diff
@@ -21,44 +23,23 @@ void kfence_cache_unregister(struct kmem_cache *s);
...
-{
-}
-static inline void *kfence_alloc_and_fix_freelist(struct kmem_cache *s)
-{
-	return NULL;
-}
+#else /* CONFIG_KFENCE */
+
+static inline void kfence_init(void) { }
+static inline void *kfence_alloc_and_fix_freelist(struct kmem_cache *s) { return NULL; }
```
ramosian-glider, line 33: Did clang-format do this change? Otherwise we'll lose the formatting next time we apply the tool.

lib/Kconfig.debug
```diff
@@ -2163,7 +2163,8 @@ config TEST_MEMINIT
 	  If unsure, say N.
 
 config TEST_KFENCE
-	tristate "Test KFENCE"
+	tristate "Test KFence"
```
ramosian-glider, line 2166: I like "KFence" more than "KFENCE", thanks!

mm/Kconfig.debug
```diff
@@ -140,8 +140,12 @@ config PTDUMP_DEBUGFS
 	  If in doubt, say N.
 
 config KFENCE
-	bool "KFence"
-	#depends on DEBUG_KERNEL
+	bool "KFence: low-overhead use-after-free and buffer-overflow detector"
+	depends on !KASAN
```
ramosian-glider: A comment about why we disable KASAN would be nice.

include/linux/slab.h
```diff
@@ -113,10 +113,9 @@
 #endif
 
 #ifdef CONFIG_KFENCE
-#ifdef CONFIG_KASAN
-#error Conflicting definitions
-#endif
 #define SLAB_KFENCE		((slab_flags_t __force)0x08000000U)
```
ramosian-glider, line 116: We probably need a comment that this flag clashes with SLAB_KASAN.

mm/Kconfig.debug
```diff
@@ -140,8 +140,12 @@ config PTDUMP_DEBUGFS
...
 	select STACKDEPOT
-	---help---
-	  Nothing here
+	help
+	  KFence is low-overhead sampling-based detector for heap
+	  use-after-free and buffer-overflow errors. KFence is designed to have
+	  negligible cost to permit enabling it in production environments.
+
+	  See <file:Documentation/vm/kfence.rst> for more details.
```
ramosian-glider, line 154: We don't have this file yet, correct?

include/linux/slab.h
melver: Can add comment. Kconfig 'depends on !KASAN' takes care of enforcing this though.

mm/Kconfig.debug
melver: Will add.

mm/Kconfig.debug
melver: No it's not there yet, but it should be in the future. ;-)

include/linux/kfence.h
melver: No it didn't. This is what the formatting should look like for these in the end. I think it's not too much of an issue to manually format the header file. Alternatively we could put

// clang-format off

here, and remove when we're done. I'll go with // clang-format off.

arch/x86/Kconfig
```diff
@@ -138,6 +138,7 @@ config X86
 	select HAVE_ARCH_JUMP_LABEL_RELATIVE
 	select HAVE_ARCH_KASAN			if X86_64
 	select HAVE_ARCH_KASAN_VMALLOC		if X86_64
+	select HAVE_ARCH_KFENCE
```
ramosian-glider, line 141: Hm, wonder how this worked before.

include/linux/slab.h
ramosian-glider: That's true, but people looking at slab.h may not know KFENCE doesn't work together with KASAN

arch/x86/Kconfig
melver: config KFENCE just didn't depend on this. I've added it together with this.

PR #67, opened 2020-05-06
kfence: factor out code that dumps objects to a buffer

ramosian-glider:
We'll be using it for the debugfs handler


ramosian-glider:
Releasing the yesterday's patch while I am stuck with debugfs :)

PR #68, opened 2020-05-07
Implement /debugfs/kfence/objects

ramosian-glider:


mm/kfence.c
```diff
@@ -474,8 +474,8 @@ static int kfence_dump_object(char *buf, size_t buf_size, int obj_index,
 	len += snprintf(buf + len, buf_size - len, "allocated at:\n");
 	len += kfence_dump_stack(buf + len, buf_size - len, obj, true);
 	if (kfence_metadata[obj_index].state == KFENCE_OBJECT_FREED) {
-		len = snprintf(buf + len, buf_size - len, "freed at:\n");
```
melver, line 477: Sorry for not spotting this earlier:

s/snprintf/scnprintf/

snprintf() returns what would be generated, regardless if the string is truncated due to size limit or not. scnprintf() returns what we'd expect.

Also, for strings that do not contain any formatting, why not use one of str[nls]cpy? strscpy() should work, but will return -E2BIG if not everything was copied.

ramosian-glider: Indeed you are right: https://lwn.net/Articles/69419/
But what happens if we do `len += scnprintf(...)`, and the latter returns -E2BIG?

melver: scnprintf() doesn't return -E2BIG, only strscpy(). If scnprintf() writes 0 bytes, it returns 0.

ramosian-glider: Got it. I've added a patch that replaces `snprintf()` with `scnprintf()`

Issue #62, opened 2020-05-02
[kfence] Stolen freelists are CPU-local


ramosian-glider:
SLUB uses per-CPU freelists on allocation fast path, so we need to ensure we steal evenly from each CPU.
`kfence_steal_freelist()` always steals the freelist from the current CPU, and `kfence_fix_freelist()`fixes it for the current CPU, even if the freelist was stolen from another CPU.
We may want to do one of the following:

- steal every `cache->cpu_slab->freelist` for the cache in `kfence_steal_freelist()`. In that case every allocation from that cache on every other CPU will be waiting for `kfence_alloc_and_fix_freelist()` to finish and restore the freelists.
- pick a random CPU in `kfence_steal_freelist()` and steal only the corresponding freelist.

melver:
I think the fairest result will be obtained by picking a random CPU. Fair as in on every system, regardless of number of CPUs, adjusting the sample rate has a predictable result on system behaviour without having to think about number of CPUs and factoring that into what the sample rate should be.

ramosian-glider:
Should be fine now.

PR #70, opened 2020-05-07
Fix issue #62 and data races on freelist pointers

ramosian-glider:


mm/kfence.c
```diff
@@ -733,6 +735,7 @@ void kfence_observe_memcg_cache(struct kmem_cache *memcg_cache)
 	unsigned long flags;
 	struct kmem_cache *root;
 	char *name;
+	int cpu = smp_processor_id();
```
melver, line 745: If some random CPU, where the current one is the most likely one, is what you want, need to use raw_smp_processor_id().

Otherwise move it inside the IRQ-off region where it's used.

mm/kfence.c
```diff
@@ -710,12 +711,18 @@ static void kfence_steal_freelist(struct kmem_cache *cache, int cpu)
...
-	c->freelist = 0;
+	/*
+	 * We need to atomically read the old value from c->freelist and write
+	 * NULL to it, but SLUB may allocate from this CPU cache and replace
+	 * c->freelist in the meantime. Use cmpxchg loop to ensure fl->freelist
+	 * contains the latest value.
+	 */
+	do {
+		fl->freelist = READ_ONCE(c->freelist);
+	} while (cmpxchg(&c->freelist, fl->freelist, NULL) != fl->freelist);
```
melver, line 725: We should just test KFence with KCSAN at some point. Not urgent right now, but eventually. I'll create an issue.

mm/kfence.c
ramosian-glider: Thanks!

PR #75, opened 2020-05-21
kfence: the great schism

ramosian-glider:
Split kfence.c into kfence_core.c (pool allocator, error reporting,
debugfs) and kfence_steal.c (heartbeat timer and freelist stealing
mechanism).
This will ease trying other sampling approaches, e.g. the naive
counter-based one. Those implementations will need to implement their own
kfence_init() and use the API declared in kfence_core.h.

The change also helped discover an incorrect lock having been used in
obj_show().


melver:
Nice, thanks!

Does it make sense to move to mm/kfence/ at this point?

ramosian-glider:
Yes, probably. I thought for two files it didn't make sense, but that was before `kfence_core.h`. I'll create the dir as well.

PR #76, opened 2020-05-26
More refactorings, add CONFIG_KFENCE_NAIVE

ramosian-glider:
I don't want to dive into benchmarking right now, but an alternative sampling algorithm will help discover more bugs in the core.

### DIFF
--- a/mm/kfence/kfence_core.c
+++ b/mm/kfence/kfence_core.c
@@ -50,8 +50,8 @@ static unsigned long kfence_pool_start, kfence_pool_end;
 /* Protects kfence_freelist, kfence_recycle, kfence_metadata */
 static DEFINE_SPINLOCK(kfence_alloc_lock);
 
-#define KFENCE_DUMP_BUF_SIZE (PAGE_SIZE * 2)
-static char kfence_dump_buf[KFENCE_DUMP_BUF_SIZE];
+/* Size picked to accommodate the metadata of a single KFENCE object. */
+static char kfence_dump_buf[PAGE_SIZE * 2];
 
 /*
  * kfence_freelist is a wrapper around kfence page pointers that allows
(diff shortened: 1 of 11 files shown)


mm/kfence/kfence_core.c
```diff
@@ -259,14 +259,14 @@ static inline unsigned long kfence_index_to_addr(int index)
 	return kfence_obj_to_addr(obj, index);
 }
 
-void *guarded_alloc(struct kmem_cache *cache, gfp_t gfp)
+void *guarded_alloc(struct kmem_cache *cache, size_t override_size, gfp_t gfp)
```
melver, line 263: Sorry, haven't noticed before: I think since these are global names, they will require a kfence_ prefix.

mm/kfence/kfence_core.c
```diff
@@ -50,6 +50,9 @@ static unsigned long kfence_pool_start, kfence_pool_end;
 /* Protects kfence_freelist, kfence_recycle, kfence_metadata */
 static DEFINE_SPINLOCK(kfence_alloc_lock);
 
+#define KFENCE_DUMP_BUF_SIZE (PAGE_SIZE * 2)
```
melver, line 53: Since this is private to this compilation unit, why not save us a macro and just have:

/* Size is chosen to accommodate ...?? */
static char kfence_dump_buf[PAGE_SIZE * 2];

and below just use sizeof(kfence_dump_buf).

mm/kfence/kfence_core.c
ramosian-glider: Right, I just forgot to rename it.

mm/kfence/kfence_core.c
ramosian-glider: Yes, good idea.

PR #77, opened 2020-06-05
Refactor initialization + minor fixes

ramosian-glider:
Ensure that KFENCE starts only after the RNG is working.


melver:
Looks good, thanks!

Issue #79, opened 2020-06-16
[kfence] Use patterns to detect non-faulting OOB writes


ramosian-glider:
Because not all OOB accesses to KFENCE objects result in page faults, we can use pattern initialization as a best-effort protection measure: initialize the whole page with a pattern when allocating an object and perform an integrity check of unused parts when freeing it.

ramosian-glider:
Got the first results from a nightly syzkaller run:

```
==================================================================
BUG: KFENCE: memory corruption at address ffff88813b370f03 on object #183 
CPU: 0 PID: 5 Comm: kworker/0:0 Not tainted 5.8.0-rc1+ #381 
Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.13.0-1 04/01/2014
Workqueue: events kfree_rcu_work
Call Trace:
 __dump_stack lib/dump_stack.c:77
 dump_stack+0xd4/0x13e lib/dump_stack.c:118
 kfence_report_corruption mm/kfence/core.c:467
 set_or_check_canary_byte.cold+0x73/0xa0 mm/kfence/core.c:268
 set_or_check_canaries+0xc2/0xe0 mm/kfence/core.c:282
 kfence_guarded_free+0x131/0x1d0 mm/kfence/core.c:356
 kfence_free+0xac/0x100 mm/kfence/core.c:377
 __slab_free+0x31/0x520 mm/slub.c:2927
 do_slab_free mm/slub.c:3082
 slab_free mm/slub.c:3095
 kmem_cache_free_bulk mm/slub.c:3219
 kmem_cache_free_bulk+0x3bb/0x460 mm/slub.c:3207
 kfree_bulk ./include/linux/slab.h:413
 kfree_rcu_work+0xf5/0x3e0 kernel/rcu/tree.c:3051
 process_one_work+0x38a/0x900 kernel/workqueue.c:2269
 worker_thread+0x56/0x5c0 kernel/workqueue.c:2415
 kthread+0x17b/0x1c0 kernel/kthread.c:291
 ret_from_fork+0x1f/0x30 arch/x86/entry/entry_64.S:293
Object #183: starts at ffff88813b370e00, size=259
allocated at:
 save_stack+0x33/0x70 mm/kfence/core.c:95
 kfence_guarded_alloc+0x28a/0x3e0 mm/kfence/core.c:327
 kfence_alloc_with_size mm/kfence/naive.c:26
[...]

PR #81, opened 2020-06-19
Use canaries to detect non-fatal OOB writes.

ramosian-glider:
See https://github.com/google/kasan/issues/79

### DIFF
--- a/lib/test_kfence.c
+++ b/lib/test_kfence.c
@@ -117,7 +117,18 @@ static int do_test_oob(size_t size, bool use_cache)
 	return res;
 }
 
-static int do_test_kmalloc73_oob(void)
+/*
+ * In the following situation:
+ *   char *p = kmalloc(73, GFP_KERNEL);
+ *   READ_ONCE(p[73]);
+ * @p is aligned on 8 (alignment of kmalloc-96), therefore the allocated object
+ * does not adhere to either of the page boundaries. Therefore an immediate
+ * buffer overflow won't trigger a page fault.
+ *
+ * This test checks that KFENCE is unable to detect such OOBs, but is able to
+ * detect an OOB that touches the next 8 bytes past the object.
+ */
+static int do_test_kmalloc_aligned_oob_read(void)
 {
 	void *buffer;
 	char *c;
@@ -198,7 +209,7 @@ static int __init test_kfence_init(void)
 
 	failures += do_test_oob(32, false);
 	failures += do_test_oob(32, true);
-	failures += do_test_kmalloc73_oob();
+	failures += do_test_kmalloc_aligned_oob_read();
 	failures += do_test_uaf(32, false);
 	failures += do_test_uaf(32, true);
 	failures += do_test_shrink(32);
--- a/lib/test_kfence.c
+++ b/lib/test_kfence.c
@@ -162,6 +162,30 @@ static int do_test_kmalloc_aligned_oob_read(void)
 	return res;
 }
 
+static int do_test_kmalloc_aligned_oob_write(void)
+{
+	void *buffer;
+	unsigned char *c, value;
+	int res = 0;
+	const size_t size = 73;
+
+	buffer = alloc_from_kfence(size, GFP_KERNEL, SIDE_RIGHT, __func__);
+	if (buffer) {
+		/*
+		 * The object is offset to the right, so we won't get a page
+		 * fault immediately after it.
+		 */
+		c = ((char *)buffer) + size + 1;
+		value = READ_ONCE(*c);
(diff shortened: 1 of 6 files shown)


mm/kfence/core.c
```diff
@@ -255,6 +255,32 @@ size_t kfence_ksize(const void *addr)
 	return abs(READ_ONCE(kfence_metadata[index].size));
 }
 
+static void kfence_report_corruption(unsigned long address);
+
+void set_or_check_canary_byte(unsigned long addr, bool set)
```
melver, line 260: static?

mm/kfence/core.c
```diff
@@ -255,6 +255,32 @@ size_t kfence_ksize(const void *addr)
 	return abs(READ_ONCE(kfence_metadata[index].size));
 }
 
+static void kfence_report_corruption(unsigned long address);
```
melver, line 258: Can they be moved so this it not required?

mm/kfence/core.c
```diff
@@ -255,6 +255,32 @@ size_t kfence_ksize(const void *addr)
...
+	char p = pattern[addr % ARRAY_SIZE(pattern)];
+	if (set) {
+		*(char *)addr = p;
+	} else {
+		if (*(char *)addr != p)
+			kfence_report_corruption(addr);
+	}
+}
+
+void set_or_check_canaries(int index, bool set)
```
melver, line 272: static?

mm/kfence/core.c
```diff
@@ -308,7 +314,7 @@ void *kfence_guarded_alloc(struct kmem_cache *cache, size_t override_size,
 		kfence_metadata[index].alloc_stack =
 			save_stack(gfp & ~__GFP_RECLAIM);
 		kfence_metadata[index].cache = cache;
-		kfence_metadata[index].size = right ? -size : size;
+		WRITE_ONCE(kfence_metadata[index].size, right ? -size : size);
```
melver, line 328: Just FYI, we can now run KCSAN on KFENCE. ;-)

This is more an AI for me, when I get back to reworking the test into more of a torture test.

lib/test_kfence.c
```diff
@@ -162,6 +162,30 @@ static int do_test_kmalloc_aligned_oob_read(void)
...
 
+static int do_test_kmalloc_aligned_oob_write(void)
+{
+	void *buffer;
+	unsigned char *c, value;
+	int res = 0;
+	const size_t size = 73;
+
+	buffer = alloc_from_kfence(size, GFP_KERNEL, SIDE_RIGHT, __func__);
+	if (buffer) {
```
melver, line 173: if (!buffer) return 1; 
... unindented rest...

?

mm/kfence/core.c
melver, line 272: Can we make this 'for_each_canary(int index, void (* fn)(unsigned long))' ?

If we want performance, we'd likely have to maco'ify this anyway. But I don't think this is performance critical.

mm/kfence/core.c
ramosian-glider: ack

mm/kfence/core.c
ramosian-glider: done

mm/kfence/core.c
ramosian-glider: done

ramosian-glider: Not sure I get it. What is `fn`?

I also think our compilers are smart enough to inline `set_or_check_canary_byte()` here (and even inline `set_or_check_canaries()` in the places where `set` is a constant.

lib/test_kfence.c
ramosian-glider: Fixed here and in the previous test.

mm/kfence/core.c
ramosian-glider: These reporting functions actually belong together, I don't think it'll look better if we move one of them to the top of the file.
I've been thinking about moving them away to `mm/kfence/report.c`, but that's not too much code now, so I'll just move this declaration to `core.h`

mm/kfence/core.c
melver: I was only suggesting getting rid of the ifs in there, and replacing it with a function pointer.

If that's better or not, I don't know. It seems more readable, but if performance is a concern, maybe the compiler isn't clever enough, unless you make for_each_canary __always_inline I think.

ramosian-glider: Done. On the other hand, we could make the second parameter of `for_each_canary` an enum with actions instead of a function ptr. We can change it if we notice that being a performance problem.

PR #84, opened 2020-06-26
Cleanups and refactors

melver:
Hi Alex,

Some cleanups and refactors. More to come, but I wanted to get this sent now in case you are working on conflicting changes.

I still need to figure out how to minimize the externs and non-statics we now have, but hopefully those should be smaller changes, rather than moving entire functions around.

If you're not working on conflicting changes right now, feel free to defer review+merge.

Thanks,
-- Marco


mm/kfence/naive.c
```diff
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: GPL-2.0
-//
```
ramosian-glider, line 2: This piece is unrelated, but I don't mind.

melver: Yeah, I'm going to have some of these minor fixes all over the place. There is no point in creating separate commits -- we're not sending these patches to LKML. ;-)

ramosian-glider:
Thanks!

PR #87, opened 2020-07-01
Use a single inline function for sampled allocations

ramosian-glider:
Use a per-CPU sample counter, adding a single load and a branch to the fast path.
kmalloc() requires some plumbing - we pass the requested size via `_RET_IP_` now.
This might be a bit questionable, but ends up being almost non-intrusive.

### DIFF
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -2,6 +2,7 @@
 #ifndef _LINUX_KFENCE_H
 #define _LINUX_KFENCE_H
 
+#include <linux/percpu.h>
 #include <linux/types.h>
 
 struct kmem_cache;
@@ -68,7 +69,23 @@ static inline void kfence_observe_memcg_cache(struct kmem_cache *memcg_cache) { 
 
 #ifdef CONFIG_KFENCE_NAIVE
 
+DECLARE_PER_CPU(int, kfence_sample_cnt);
+extern unsigned long kfence_sample_rate;
+
 void *kfence_alloc_with_size(struct kmem_cache *s, size_t size, gfp_t flags);
+static __always_inline void *kfence_alloc_with_addr(struct kmem_cache *s, gfp_t flags, unsigned long addr)
+{
+	int cnt = this_cpu_dec_return(kfence_sample_cnt);
+	size_t size;
+
+	if (likely(cnt > 0))
+		return NULL;
+	this_cpu_write(kfence_sample_cnt, kfence_sample_rate);
+	size = addr >> 48;
+	if (size == 0xffff)
+		size = 0;
+	return kfence_alloc_with_size(s, size, flags);
+}
 
 #else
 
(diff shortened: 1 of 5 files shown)


include/linux/kfence.h
```diff
@@ -68,14 +69,31 @@ static inline void kfence_observe_memcg_cache(struct kmem_cache *memcg_cache) {
...
 void *kfence_alloc_with_size(struct kmem_cache *s, size_t size, gfp_t flags);
+static __always_inline void *kfence_alloc_with_addr(struct kmem_cache *s, gfp_t flags, unsigned long addr)
+{
+	int cnt = this_cpu_dec_return(kfence_sample_cnt);
+	size_t size;
+
+	if (likely(cnt > 0))
+		return NULL;
+	this_cpu_write(kfence_sample_cnt, kfence_sample_rate);
+	size = addr >> 48;
```
melver, line 84: This assume that 'addr' is 64bit, which I don't think we can do.

You could calculate this as 'BITS_PER_LONG - 16'.

mm/slub.c
```diff
@@ -2583,6 +2583,20 @@ static inline void *get_freelist(struct kmem_cache *s, struct page *page)
...
 }
 
+/*
+ * If using KFENCE, allow passing object size in the top 16 bits of _RET_IP_,
+ * assuming they are set to 0xFFFF otherwise (currently only 64-bit systems).
+ */
+#ifdef CONFIG_KFENCE
+#define RET_IP_SIZE(size) \
+	((((unsigned long)(_RET_IP_) << 16) >> 16) | \
+	 ((unsigned long)size << 48))
```
melver, line 2593: I think this needs to be calculated in terms of BITS_PER_LONG.

mm/slub.c
```diff
@@ -3925,11 +3944,7 @@ void *__kmalloc(size_t size, gfp_t flags)
 	if (unlikely(ZERO_OR_NULL_PTR(s)))
 		return s;
 
-	ret = kfence_alloc_with_size(s, size, flags);
-	if (ret)
-		return ret;
-
-	ret = slab_alloc(s, flags, _RET_IP_);
+	ret = slab_alloc(s, flags, RET_IP_SIZE(size));
```
melver, line 3947: I think this will never pass upstream code review.

Both slab_alloc_node and slab_alloc are __always_inline. So any additional parameter, if unused, will be optimized out.

So, I think just adding another argument with size and then just passing 'size' instead of 'addr' to KFENCE is better. It also removes the bit-twiddling.

ramosian-glider: Hm, you are right, `_RET_IP_` is only passed to __slab_alloc, otherwise it just dissolves.
Adding another argument sure looks cleaner.

mm/slub.c
ramosian-glider: Noted, but I'll be dropping this patch now.

include/linux/kfence.h
ramosian-glider: Ack.

mm/kfence/report.c
```diff
@@ -105,6 +105,22 @@ static void kfence_print_object(int obj_index,
...
+static void dump_bytes_at(unsigned long addr)
+{
+	unsigned char *c = (unsigned char *)addr;
+	unsigned char *max_addr = (unsigned char *)min(ALIGN(addr, PAGE_SIZE),
+						       addr + BYTES_TO_DUMP);
+	char bytes[BYTES_TO_DUMP * 3 + 1];
+	int len = 0;
+
+	for (; c < max_addr; c++)
+		len += scnprintf(bytes + len, sizeof(bytes) - len, "%02X ", *c);
```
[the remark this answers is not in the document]
ramosian-glider: Thanks for reminding me of pr_cont()!

PR #88, opened 2020-07-02
Rework https://github.com/google/kasan/pull/87

ramosian-glider:


mm/kfence/naive.c
```diff
@@ -46,6 +41,6 @@ void kfence_impl_init(void)
 	 * TODO: a better idea would be to use a normal distribution around
 	 * kfence_sample_rate.
 	 */
-	for_each_possible_cpu(i)
```
melver, line 49: clang-format having a mind of its own?

mm/slub.c
```diff
@@ -2753,13 +2754,17 @@ static __always_inline void maybe_wipe_obj_freeptr(struct kmem_cache *s,
  * Otherwise we can simply pick the next object from the lockless free list.
  */
 static __always_inline void *slab_alloc_node(struct kmem_cache *s,
-		gfp_t gfpflags, int node, unsigned long addr)
+		gfp_t gfpflags, int node, unsigned long addr, size_t cust_size)
```
melver, line 2757: I assume we somewhere clash with 'size'?

If 'cust_size' is the original size, why not 'orig_size'? Or 'caller_size'?

ramosian-glider: Perhaps not in this function, but clashes are indeed possible.
"orig_size" looks good.

PR #89, opened 2020-07-03
kfence: Rework object printing

melver:
Remove kfence_dump_buf and rework how objects are printed to either
seq_files or to console.


mm/kfence/report.c
```diff
@@ -1,13 +1,29 @@
 // SPDX-License-Identifier: GPL-2.0
 
+#include <stdarg.h>
```
ramosian-glider, line 3: Hm, I didn't know we can use stdarg.h

PR #92, opened 2020-07-09
Support constructors

ramosian-glider:


mm/kfence/static_key.c
```diff
@@ -26,7 +26,8 @@ void *kfence_alloc_with_size(struct kmem_cache *s, size_t size, gfp_t flags)
 	 * sense to continue writing to it and pay the associated contention
 	 * cost, in case we have a large number of concurrent allocations.
 	 */
-	if (atomic_read(&allocation_gate) || atomic_inc_return(&allocation_gate) > 1)
```
melver, line 29: I don't mind either way now, because we'll redo a bunch of these anyway.

But, what we have now is what the final version should be, because the kernel officially changed permitted column width to 100 chars. It's a lot more readable if it's all on the same line.

ramosian-glider: Fine, I'll drop this change and fix .clang-format in our tree then.

Issue #66, opened 2020-05-05
[kfence] Reconsider using stack depot


ramosian-glider:
It is nice that we don't have to invent our own storage for stack traces, but stackdepot may increase our memory footprint, as it does not garbage collect stacks.

It is actually enough to allocate `KFENCE_STACK_DEPTH * 8 bytes * 2 for alloc/free stacks * 256 objects` bytes once and overwrite old stacks as objects get reused.

dvyukov:
I agree, I think we should not use stack depot for KFENCE.

PR #94, opened 2020-07-09
kfence: get rid of stack depot

ramosian-glider:
Unlike other debugging tools, we have a fixed number of allocations to
care about, so we'd better store the stacks ourselves to avoid the
growing memory consumption of stackdepot.

Fixes https://github.com/google/kasan/issues/66


dvyukov:
+1

PR #95, opened 2020-07-15
kfence: statically allocate the memory pool

ramosian-glider:
Making __kfence_pool_start an array in the data section allows us to
check whether an address belongs to it or not without doing memory
loads.

This also required fixing kfence_force_4k_pages(), which didn't work
initially.

### DIFF
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 
+#include <asm/pgalloc.h>
+
 #include <linux/debugfs.h>
 #include <linux/kfence.h>
 #include <linux/list.h>
@@ -23,7 +25,8 @@ struct kfence_freelist {
 	void *obj;
 };
 
-unsigned long __kfence_pool_start;
+char __kfence_pool_start[PAGE_SIZE << (KFENCE_NUM_OBJ_LOG + 1)]
+	__attribute__((section(".kfence"), aligned(1 << 21)));
 EXPORT_SYMBOL(__kfence_pool_start);
 
 /* Protects kfence_freelist, kfence_recycle, kfence_metadata */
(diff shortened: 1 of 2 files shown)


mm/kfence/core.c
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 
+#include <asm/pgalloc.h>
```
melver, line 3: Can this not be grouped with asm includes below? Does it pull in something needed by another include?

mm/kfence/core.c
```diff
@@ -71,31 +74,135 @@ noinline void kfence_disable(void)
...
+	 * _PAGE_PROTNONE.
+	 */
+	if (!(pgprot_val(prot) & _PAGE_PRESENT))
+		pgprot_val(prot) &= ~_PAGE_GLOBAL;
+
+	return prot;
+}
+
+/* Some code borrowed from arch/x86/mm/pat/set_memory.c. */
+static bool split_large_page(pte_t *kpte, unsigned long address, unsigned int level)
```
melver, line 94: This is all x86-specific, right? We're going to need to move this to something like '<asm/kfence.h>' eventually. A TODO would be good so we don't forget.

mm/kfence/core.c
```diff
@@ -25,8 +25,7 @@ struct kfence_freelist {
 	void *obj;
 };
 
-char __kfence_pool_start[PAGE_SIZE << (KFENCE_NUM_OBJ_LOG + 1)]
-	__attribute__((section(".kfence"), aligned(1 << 21)));
+char __kfence_pool_start[PAGE_SIZE << (KFENCE_NUM_OBJ_LOG + 1)] __attribute__((aligned(2 << 21)));
```
melver, line 28: `__aligned(...)` can be used now.

mm/kfence/core.c
ramosian-glider: I actually think page table interfaces are arch-agnostic, but the devil's in the details.
Let's keep a TODO around till we try it on ARM64.

mm/kfence/core.c
ramosian-glider: Done.

PR #96, opened 2020-07-15
Drop non-default sampling algorithms, code cleanup

ramosian-glider:


include/linux/kfence.h
```diff
@@ -19,6 +20,20 @@ struct kmem_cache;
...
 extern char __kfence_pool_start[];
+extern struct static_key_false kfence_allocation_key;
+
+// TODO(elver): Shorten function names, and make __kfence_alloc private.
+// Simply rename them to __kfence_alloc and kfence_alloc respectively.
+
+void *__kfence_alloc(struct kmem_cache *s, size_t size, gfp_t flags);
+
+// TODO(elver): Add API doc.
+static __always_inline void *kfence_sampled_alloc_with_size(struct kmem_cache *s, size_t size,
```
melver, line 31: This one was meant to be `kfence_alloc`, now that you already renamed to `__kfence_alloc` above.

I think what I meant by making "private" is adding the __ in front, which seems to be the kernel way of denoting internal things. Renaming this means you'll have resolved the above TODO. Thank you!

include/linux/kfence.h
```diff
@@ -55,86 +70,12 @@ static inline bool kfence_handle_page_fault(unsigned long addr) { return false;
...
-
-#else
-
-// TODO: remove for v1
-// clang-format off
-
-static inline void *kfence_alloc_with_size(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
-static __always_inline void *kfence_sampled_alloc_with_size(struct kmem_cache *s, gfp_t flags, size_t size) { return NULL; }
+static inline void *__kfence_alloc(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
+static __always_inline void *kfence_sampled_alloc_with_size(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
```
melver, line 74: kfence_alloc for this one then as well.

mm/kfence/core.c
```diff
@@ -372,12 +372,24 @@ static void for_each_canary(int index, void (*fn)(unsigned long))
...
+static DECLARE_WAIT_QUEUE_HEAD(allocation_wait);
+
 void *kfence_guarded_alloc(struct kmem_cache *cache, size_t override_size, gfp_t gfp)
 {
 	unsigned long flags;
 	void *obj = NULL, *ret;
 	struct kfence_freelist *item;
 	int index = -1;
+	/*
+	 * TODO(glider): for allocations made before RNG initialization prandom_u32_max() will
```
melver, line 390: I think this doesn't need to be a TODO. We can just 'Note: ' this, because if we do not address it for v1, I don't think anybody will care. We can motivate this also by adding "We still benefit from enabling KFENCE as early as possible, even when the RNG is not yet available, as this will allow KFENCE to detect bugs due to earlier allocations. The only downside is that the out-of-bounds accesses detected are deterministic for such allocations."

include/linux/kfence.h
ramosian-glider: Yes, I forgot to fix that one. Will do.

mm/kfence/core.c
ramosian-glider: Ok.

include/linux/kfence.h
ramosian-glider: Done.

PR #99, opened 2020-07-21
Update some documentation and MAINTAINERS

melver:


Documentation/dev-tools/kfence.rst
```diff
@@ -0,0 +1,48 @@
+Kernel Electric-Fence
+=====================
+
+Kernel Electric-Fence (KFENCE) is a low-overhead sampling-based memory
+debugger. KFENCE detects kmalloc use-after-free and buffer-overflow errors.
```
ramosian-glider, line 5: I think "heap" is better than "kmalloc", as there are heap allocations that aren't kmalloc allocations.

melver: Done.

PR #100, opened 2020-07-21
Implement SLAB support

ramosian-glider:


mm/kfence/core.c
```diff
@@ -348,7 +348,11 @@ static void check_cache_freelist_ptr(int index)
 
 	if (!meta->cache)
 		return;
+#if defined(CONFIG_SLUB)
```
melver, line 351: We should avoid #ifdef 

```
if (IS_ENABLED(CONFIG_SLAB))
  return;
```

Our Kconfig should take care of ensuring we don't have something other than SL[AU]B, right?

Otherwise we might need a

```
BUILD_BUG_ON(IS_ENABLED(CONFIG_SLUB) || IS_ENABLED(CONFIG_SLAB));
```

include/linux/kfence.h
```diff
@@ -50,6 +38,24 @@ static inline bool is_kfence_addr(void *addr)
...
+
+bool __kfence_free(void *addr);
+
+static __always_inline bool kfence_free(void *addr)
+{
+	if (!is_kfence_addr(addr))
+		return false;
+	return __kfence_free(addr);
+}
+
```
melver, line 58: I think you tried to reorder to make consistent with what's in the '#else ' (noops), right? Or another reason this moved?

kfence_ksize is still in the wrong place in that case relative to these. I'd probably also move kfence_free after kfence_alloc (here and below in #else).

mm/slab.c
```diff
@@ -3481,7 +3495,7 @@ void ___cache_free(struct kmem_cache *cachep, void *objp,
  */
 void *kmem_cache_alloc(struct kmem_cache *cachep, gfp_t flags)
 {
-	void *ret = slab_alloc(cachep, flags, _RET_IP_);
+	void *ret = slab_alloc(cachep, flags, 0, _RET_IP_);
```
melver, line 3498: Could we just use cachep->size? That way we'd be explicit about it, because it'll be confusing to reviewers trying to understand this.

I see that kfence_guarded_alloc still has the 'size = override_size ? ...' so this works, but I think if we can get rid of that hack would be better. I think we don't not know the real size anywhere now, right?

Then, in kfence_guarded_alloc, we can have a KFENCE_WARN or so if the size is 0. Or we could declare it a real error. https://lwn.net/Articles/236920/ seems to indicate it's a stupid thing to do and  slub/slab return ZERO_SIZE_PTR if someone decided to do a kmalloc(0).

mm/slab.c
```diff
@@ -3572,7 +3586,7 @@ EXPORT_SYMBOL(kmem_cache_alloc_trace);
  */
 void *kmem_cache_alloc_node(struct kmem_cache *cachep, gfp_t flags, int nodeid)
 {
-	void *ret = slab_alloc_node(cachep, flags, nodeid, _RET_IP_);
+	void *ret = slab_alloc_node(cachep, flags, nodeid, 0, _RET_IP_);
```
melver, line 3589: Same here

mm/kfence/core.c
ramosian-glider: > We should avoid #ifdef

In this case we cannot. SLAB has no meta->cache->offset, so the code won't compile.
OTOH I think it's cleaner to just #ifdef the whole function body (see the updated patch).

ramosian-glider: > Our Kconfig should take care of ensuring we don't have something other than SL[AU]B, right?

Added this, thanks!

include/linux/kfence.h
ramosian-glider: No, the reason was to put kfence_free() after is_kfence_addr() :)
I've reordered the functions in the noop section for consistency.

mm/slab.c
ramosian-glider: I think you are right.
There's a check for !cachep in slab_alloc_node(), but it is dead already, as cachep is being dereferenced in the callers of `slab_alloc_node()`.

mm/slab.c
ramosian-glider: Done

PR #101, opened 2020-07-23
Cleanups and improve reporting

melver:
Some cleanups and reporting changes in preparation of the test rewrite using KUnit.

Before I go and finalize the test, I'd like to finalize the format of the reports.

### DIFF
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -68,9 +68,7 @@ static inline bool kfence_discard_slab(struct kmem_cache *s, struct page *page) 
 static inline bool kfence_handle_page_fault(unsigned long addr) { return false; }
 static inline bool is_kfence_addr(void *addr) { return false; }
 static __always_inline void *kfence_alloc(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
-static inline bool kfence_free(struct kmem_cache *s, struct page *page,
-			       void *head, void *tail, int cnt,
-			       unsigned long addr) { return false; }
+static inline bool kfence_free(void *addr) { return false; }
 static inline size_t kfence_ksize(void *addr) { return 0; }
 
 // TODO: remove for v1
--- a/lib/Kconfig.kfence
+++ b/lib/Kconfig.kfence
@@ -22,3 +22,15 @@ config KFENCE
 	  such as KASAN. KFENCE can detect a subset of bugs that KASAN can
 	  detect (therefore enabling KFENCE together with KASAN does not make
 	  sense), albeit at very different performance profiles.
+
+if KFENCE
+
+config KFENCE_SAMPLE_RATE
+	int "KFENCE default sample rate in milliseconds"
+	default 100
+	help
+	  The KFENCE sample rate determines the frequency with which heap
+	  allocations will be guarded by KFENCE. May be overridden via
+	  command-line argument kfence.sample_rate.
+
+endif # KFENCE
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -14,7 +14,6 @@
 #include <asm/tlbflush.h>
 
 #include "kfence.h"
-#include "../slab.h"
 
 unsigned long kfence_sample_rate = CONFIG_KFENCE_SAMPLE_RATE;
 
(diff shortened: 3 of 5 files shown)


include/linux/kfence.h
```diff
@@ -68,9 +68,7 @@ static inline bool kfence_discard_slab(struct kmem_cache *s, struct page *page)
 static inline bool kfence_handle_page_fault(unsigned long addr) { return false; }
 static inline bool is_kfence_addr(void *addr) { return false; }
 static __always_inline void *kfence_alloc(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
-static inline bool kfence_free(struct kmem_cache *s, struct page *page,
-			       void *head, void *tail, int cnt,
-			       unsigned long addr) { return false; }
+static inline bool kfence_free(void *addr) { return false; }
```
ramosian-glider, line 71: Nice catch!
What do you think about inline vs. __always_inline in our code?
I don't see much difference in practice (though some LWN sources claimed not all compilers respect `inline`), but we probably need to pick one for consistency.

lib/Kconfig.kfence
```diff
@@ -22,3 +22,15 @@ config KFENCE
 	  such as KASAN. KFENCE can detect a subset of bugs that KASAN can
 	  detect (therefore enabling KFENCE together with KASAN does not make
 	  sense), albeit at very different performance profiles.
+
+if KFENCE
+
+config KFENCE_SAMPLE_RATE
+	int "KFENCE default sample rate in milliseconds"
+	default 100
```
ramosian-glider, line 30: Sounds good. Android ppl may want to set the rate in .config rather than boot parameters.

mm/kfence/report.c
```diff
@@ -114,10 +116,11 @@ static void dump_bytes_at(unsigned long addr)
 {
 	unsigned char *c = (unsigned char *)addr;
 	unsigned char *max_addr = (unsigned char *)min(ALIGN(addr, PAGE_SIZE), addr + 16);
-	pr_err("Bytes at %px:", (void *)addr);
+
+	pr_cont("[");
 	for (; c < max_addr; c++)
 		pr_cont(" %02X", *c);
-	pr_cont("\n");
+	pr_cont(" ]");
```
ramosian-glider, line 123: Shall we bother printing the expected bytes as well?
I think people won't remember our pattern by heart (we may as well want to randomize it in the future).

mm/kfence/core.c
```diff
@@ -534,9 +531,8 @@ bool kfence_handle_page_fault(unsigned long addr)
 			}
 		}
 		if (report_index != -1) {
-			object = kfence_metadata[report_index];
+			kfence_report_error(addr, &kfence_metadata[report_index], KFENCE_ERROR_OOB);
```
ramosian-glider, line 536: Can we guarantee that kfence_report_error() won't allocate memory, causing a deadlock on kfence_alloc_lock?
This is why we've been passing the copy of kfence_metadata[report_index] to kfence_report_error().

ramosian-glider:
None of my comments is a blocker, I'll be merging this PR.

include/linux/kfence.h
melver: Right, we need to also be consistent with other kernel code.

The rule I've been following with KCSAN at least (after some iterations) was:
- Use __always_inline if it's a bug if it is not inlined (i.e. __always_inline is the "better macro"). This can be a performance bug but also things like uaccess warnings that might appear if not inlined.
- Use inline by default. The compiler might know better.
- For empty stubs, use 'inline' because any compiler that doesn't throw away empty inline functions has a bug.

For the stubs we should use 'inline' which is also consistent with other stubs across the kernel. I'll add a commit to fix.

mm/kfence/core.c
melver: Yes I think so. printk should never allocate memory (otherwise a lot more places in the kernel might be broken), and we don't use anything else in there that might.

melver:
> None of my comments is a blocker, I'll be merging this PR.

Thanks. I have some fixes to inline and the byte-dumping comments. Will push separately.

PR #104, opened 2020-07-24
Fix #102

### LINKED ISSUE #102: [kfence] Handle cache destruction
Currently we do not properly handle the case when we destroy a memcache, and still have metadata of a KFENCE object.

One thing I noticed is that when inspecting /sys/kernel/debug/kfence/objects, the printing of a memcache name of a destroyed memcache results in KFENCE doing either a UAF and/or accessing random kernel memory and showing that to userspace.

In a test I ran, in /sys/kerne/debug/kfence/objects, the names of a destroyed memcaches was showing garbage data, i.e. we were leaking random kernel memory.

ramosian-glider:
also minor formatting fixes

### DIFF
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -560,6 +560,34 @@ bool kfence_discard_slab(struct kmem_cache *s, struct page *page)
 	return true;
 }
 
+bool kfence_shutdown_cache(struct kmem_cache *s)
+{
+	unsigned long flags;
+	int i;
+	struct kfence_alloc_metadata *meta;
+	bool ret = false;
+
+	spin_lock_irqsave(&kfence_alloc_lock, flags);
+
+	for (i = 0; i < KFENCE_NUM_OBJ; i++) {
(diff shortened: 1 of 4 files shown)


mm/kfence/core.c
```diff
@@ -560,6 +560,34 @@ bool kfence_discard_slab(struct kmem_cache *s, struct page *page)
...
+		if ((meta->cache == s) && (meta->state == KFENCE_OBJECT_FREED))
+			meta->cache = NULL;
+	}
+	ret = true;
+
+leave:
+	spin_unlock_irqrestore(&kfence_alloc_lock, flags);
+	return ret;
+}
+EXPORT_SYMBOL(kfence_shutdown_cache);
```
melver, line 589: I don't think modules should use this, right? I'd expect this to only be used from allocator internals.

ramosian-glider: Ah, correct, it is called from a static function.

ramosian-glider: Fixed.

PR #105, opened 2020-07-24
Some cleanups

ramosian-glider:
Revert some leftover bits that we do not need anymore.

### DIFF
--- a/mm/slab.c
+++ b/mm/slab.c
@@ -3231,7 +3231,7 @@ slab_alloc_node(struct kmem_cache *cachep, gfp_t flags, int nodeid, size_t orig_
 
 	ptr = kfence_alloc(cachep, orig_size, flags);
 	if (ptr)
-		return ptr;
+		goto leave;
 
 	cache_alloc_debugcheck_before(cachep, flags);
 	local_irq_save(save_flags);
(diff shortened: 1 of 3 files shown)


mm/slab.c
```diff
@@ -3231,7 +3231,7 @@ slab_alloc_node(struct kmem_cache *cachep, gfp_t flags, int nodeid, size_t orig_
 
 	ptr = kfence_alloc(cachep, orig_size, flags);
 	if (ptr)
-		goto leave;
+		return ptr;
```
melver, line 3234: I think we have a problem because we executed slab_pre_alloc_hook, but now won't execute the post_alloc_hook. I think the problem might have to do with memcgs, which do something both in the pre- and post-alloc hooks.

I think we require some symmetry here, but still am not entirely sure what the right solution is: either don't execute pre-/post- hooks or execute both; but I think we must execute the pre-hook, so that really leaves only 1 option.

mm/slab.c
```diff
@@ -3315,7 +3314,7 @@ slab_alloc(struct kmem_cache *cachep, gfp_t flags, size_t orig_size, unsigned lo
 
 	objp = kfence_alloc(cachep, orig_size, flags);
 	if (objp)
-		goto leave;
+		return objp;
```
melver, line 3317: Similar here. I'm worried about the pre- and post-alloc hooks. We need to understand if we need them.

mm/slab.c
ramosian-glider: Wait, we're skipping slab_pre_alloc_hook() exactly here, so after landing this change the problem should be gone.
I can imagine however that if someone adds extra object setup to those hooks we'll have to execute them.

melver: slab_pre_alloc_hook is above at line 3228. We are skipping over slab_post_alloc_hook.

ramosian-glider: Ah, sorry, you are right. Ok, let's try to preserve both pre_ and post_alloc_hook then.

PR #107, opened 2020-07-24
KUnit test and fixes

melver:
As of this writing, one test still fails, and I'm investigating if this is a problem with the test or a real KFENCE bug. When using a custom memcache, it's not happy:

```
[   14.637354]     ok 1 - test_out_of_bounds_read                                                                              
[   14.637837]     # test_out_of_bounds_read-memcache: test_alloc: size=32, gfp=cc0, side=left, cache=1                        
[   14.744726]     # test_out_of_bounds_read-memcache: EXPECTATION FAILED at mm/kfence/kfence-test.c:236                       
[   14.744726]     Expected report_matches(&expect) to be true, but is false                                                   
[   14.747306]     # test_out_of_bounds_read-memcache: test_alloc: size=32, gfp=cc0, side=right, cache=1                       
[   14.848696]     # test_out_of_bounds_read-memcache: EXPECTATION FAILED at mm/kfence/kfence-test.c:242                       
[   14.848696]     Expected report_matches(&expect) to be true, but is false                                                   
[   14.851703]     not ok 2 - test_out_of_bounds_read-memcache 
```

Trying to figure out what's going on. Sending for early review, in case you might have ideas.

Thanks!

[Edit: above issue was fixed in kfence: Fix kmem_cache_alloc()]


mm/kfence/kfence-test.c
```diff
@@ -0,0 +1,485 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Test cases for KFENCE memory safety error detector. Since the interface with
+ * which KFENCE's reports are obtained is via the console, this is the output we
+ * should verify. For each test case checks the presence (or absence) of
+ * generated reports. Relies on 'console' tracepoint to capture reports as they
+ * appear in the kernel log.
+ *
+ * Copyright (C) 2020, Google LLC.
```
melver, line 9: I preserved this as it was in the KUnit example. I believe for modules, we're better of adding this, because modules are prone to being copied out-of-tree.

mm/kfence/kfence-test.c
```diff
@@ -0,0 +1,485 @@
...
+		}
+
+		test_free(alloc);
+	} while (jiffies < timeout);
+
+	KUNIT_ASSERT_TRUE_MSG(test, false, "failed to allocate from KFENCE");
+	return NULL; /* Unreachable. */
+}
+
+static void test_out_of_bounds_read(struct kunit *test)
```
ramosian-glider, line 258: I think we need `noinline` for the test functions.

mm/kfence/kfence-test.c
```diff
@@ -0,0 +1,485 @@
...
+	KUNIT_EXPECT_TRUE(test, report_matches(&expect));
+	test_free(buf);
+}
+
+static void test_use_after_free_read(struct kunit *test)
+{
+	const size_t size = 32;
+	struct expect_report expect = {
+		.type = KFENCE_ERROR_UAF,
+		.fn = test_use_after_free_read,
```
ramosian-glider, line 289: I couldn't come up with a macro that could be use instead of re-typing the function name for every test. Guess you also tried that?

mm/kfence/report.c
```diff
@@ -24,42 +24,35 @@ static void seq_con_printf(struct seq_file *seq, const char *fmt, ...)
...
-bool stack_entry_matches(unsigned long addr, const char *pattern)
-{
-	char buf[64];
-	int buf_len, len;
-
-	buf_len = scnprintf(buf, sizeof(buf), "%ps", (void *)addr);
-	len = strlen(pattern);
-	if (len > buf_len)
-		return false;
-	if (strnstr(buf, pattern, len))
```
ramosian-glider, line 36: Yes, this was quite silly. Thanks for catching!

mm/slab.c
```diff
@@ -3495,7 +3495,7 @@ void ___cache_free(struct kmem_cache *cachep, void *objp,
  */
 void *kmem_cache_alloc(struct kmem_cache *cachep, gfp_t flags)
 {
-	void *ret = slab_alloc(cachep, flags, cachep->size, _RET_IP_);
+	void *ret = slab_alloc(cachep, flags, cachep->object_size, _RET_IP_);
```
ramosian-glider, line 3498: Yes, for some reason I've been confusing these two sizes.

mm/kfence/kfence-test.c
melver: I thought about it, but it won't buy much -- I think it's cleaner this way. And if we mistype the function name, the test will fail anyway.

mm/kfence/kfence-test.c
melver: We pass them to KUnit as function pointers, so it will never inline any of this since the place that calls these is somewhere deep in KUnit.

mm/kfence/core.c
```diff
@@ -493,6 +493,12 @@ bool __kfence_free(void *addr)
 	save_stack(index, false);
 	kfence_metadata[index].state = KFENCE_OBJECT_FREED;
 	kfence_protect(aligned_addr);
+
+	if (kfence_metadata[index].unprotected_page) {
+		kfence_protect(kfence_metadata[index].unprotected_page);
```
ramosian-glider: Where are we going to reinstate the protection for freed pages?

lib/Kconfig.kfence
```diff
@@ -35,6 +35,14 @@ config KFENCE_NUM_OBJECTS
 	  pages are required; with one containing the object and two adjacent
 	  ones used as redzones.
 
+config KFENCE_FAULT_INJECTION
+	bool "Fault injection for stress testing"
```
ramosian-glider, line 40: Shall we make this parameter an int that we can pass to `prandom_u32_max()` instead?

mm/kfence/core.c
```diff
@@ -525,7 +527,14 @@ bool kfence_handle_page_fault(unsigned long addr)
 		return kfence_unprotect(addr);
 	}
 
+	/*
+	 * If there is a KFENCE report somewhere inside lockdep, or one of the
+	 * libraries used by it, we need to avoid recursing back into lockdep.
+	 */
+	// TODO(elver): This is probably also a problem for allocations/frees.
+	lockdep_off();
```
ramosian-glider, line 537: You are going to remove these annotations later, right?

mm/kfence/kfence-test.c
```diff
@@ -345,16 +345,20 @@ static void test_shrink_memcache(struct kunit *test)
...
-	objects[1] = kmalloc(size, GFP_KERNEL);
-	objects[2] = test_alloc(test, size, GFP_KERNEL, ALLOCATE_ANY);
-	objects[3] = kmalloc(size, GFP_KERNEL);
-	kfree_bulk(ARRAY_SIZE(objects), objects);
+	for (iter = 0; iter < 10; ++iter) {
+		const size_t size = 1 + prandom_u32_max(300);
+		void *objects[4];
 
-	KUNIT_EXPECT_FALSE(test, report_available());
+		objects[0] = test_alloc(test, size, GFP_KERNEL, ALLOCATE_RIGHT);
```
ramosian-glider: OOC, did you mean to change ALLOCATE_ANY to ALLOCATE_RIGHT/LEFT?

mm/kfence/core.c
melver: I don't think we can, because we don't know how frequent the use-after-frees are etc. For the guard pages it's obvious, we just re-protect them on free, but the use-after-freed pages it's not so obvious.

I think it's fine to just wait until the feed object is reused for another allocation, at which point its state is reset and upon free of the new object the page will be protected again.

mm/kfence/core.c
melver: Yes I fixed it already (not pushed yet).

mm/kfence/kfence-test.c
```diff
@@ -170,25 +181,44 @@ static __always_inline void test_free(void *ptr)
...
-	ALLOCATE_RIGHT,
+/*
+ * If this should be a KFENCE allocation, and on which side the allocation and
+ * the closest guard page should be.
+ */
+enum allocation_policy {
+	ALLOCATE_ANY, /* KFENCE, any side. */
+	ALLOCATE_LEFT, /* KFENCE, left side of page. */
+	ALLOCATE_RIGHT, /* KFENCE, right side of page. */
+	ALLOCATE_NONE, /* No KFENCE allocation. */
```
ramosian-glider, line 198: Nice!

mm/kfence/kfence-test.c
melver: Yes, I wanted to test both LEFT / RIGHT explicitly because I discovered some rare issue and didn't want to leave this up to randomness.

This makes the test more deterministic, and we do explicitly test both.

I think this was related to SLUB writing into the freeptr and the code that was trying to fix up the canary values. In the end I fixed it by making it never write into the redzone (somewhere in the following commits).

lib/Kconfig.kfence
melver: Yes, good idea. I'll add a commit.

mm/slub.c
```diff
@@ -3151,6 +3151,11 @@ int build_detached_freelist(struct kmem_cache *s, size_t size,
 	if (!object)
 		return 0;
 
+	if (kfence_free(object)) {
+		p[size] = NULL; /* mark object processed */
```
ramosian-glider, line 3155: Hm, I think this solves the issue with `slab_free_freelist_hook()`: now it does not need to iterate the freelist for bulk allocations.

melver: Do you mean we can remove the KFENCE check from slab_free_freelist_hook()?

mm/kfence/core.c
```diff
@@ -285,35 +254,34 @@ static bool __meminit kfence_allocate_pool(void)
 			pages[i].frozen = 1;
 		}
 	}
+
 	/* Skip the first page: it is reserved. */
+	// TODO(elver): Why is it reserved?
```
ramosian-glider: No particular reason; initially the idea was to store metadata there. Right now it's because the pool always has an even number of pages (which we could also change, actually).

melver: Yeah, thanks. I think we clarified this earlier in the meeting. I'll add a comment explaining this and remove the TODOs.

arch/x86/include/asm/kfence.h
```diff
@@ -0,0 +1,173 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+
+#ifndef _ASM_X86_KFENCE_H
```
ramosian-glider, line 3: Is "asm" a good place for this header?

mm/kfence/core.c
```diff
@@ -65,227 +63,22 @@ static atomic_t allocation_gate = ATOMIC_INIT(1);
...
-
-	return prot;
-}
-
 /*
- * Some code borrowed from arch/x86/mm/pat/set_memory.c.
- * TODO(glider): need to figure out whether this code can be used on ARM64 and change it
- * accordingly.
+ * TODO(elver): With the move of arch-specific code to asm, kfence core.c got a
+ * lot simpler. Maybe we can move report.c code back here and remove the kfence/
```
ramosian-glider, line 73: Maybe we shouldn't haste with merging the two files, because we might be adding more code in the future. Same for removing the dir.

melver: Yes, that's what I'm thinking. I'll keep it around for now. My preference is having it, but this depends on what MM maintainers consider best practice.

arch/x86/include/asm/kfence.h
melver: That's where all arch-specific code goes. Where else should it go?

mm/kfence/core.c
```diff
@@ -508,43 +512,47 @@ bool kfence_handle_page_fault(unsigned long addr)
...
+			       to_report->state != KFENCE_OBJECT_FREED);
+		error_type = KFENCE_ERROR_UAF;
 	}
 
+out:
+	if (to_report)
+		kfence_report_error(addr, to_report, error_type);
+	else
+		/* This may be a UAF or OOB access, but we can't be sure. */
+		// TODO: think about using %p everywhere to not leak kernel
```
ramosian-glider: Ugh, I hate this. For every other tool it's perfectly fine to use %px, as they are test-only.
KFENCE should really avoid printing real addresses on e.g. mobile devices, but I still think it makes sense to print them on syzkaller (should we use KFENCE there) or in server kernels.

mm/kfence/kfence.h
```diff
@@ -9,6 +9,13 @@
 
 #include "../slab.h" /* for struct kmem_cache */
 
+/* For non-debug builds, avoid leaking kernel pointers into dmesg. */
+#ifdef CONFIG_DEBUG_KERNEL
```
ramosian-glider, line 13: SGTM

mm/kfence/core.c
melver: I fixed this later. It's made conditional on CONFIG_DEBUG_KERNEL.

mm/kfence/core.c
```diff
@@ -86,20 +86,49 @@ static inline bool kfence_unprotect(unsigned long addr)
...
 		return NULL;
 
-	index = ((addr - (unsigned long)__kfence_pool) / PAGE_SIZE / 2) - 1;
+	/*
+	 * May be an invalid index if called with an address at the edge of
+	 * __kfence_pool, in which case we would report an "invalid access"
+	 * error.
+	 */
+	index = ((addr - (unsigned long)__kfence_pool) / (PAGE_SIZE * 2)) - 1;
 	if (index < 0 || index >= CONFIG_KFENCE_NUM_OBJECTS)
```
ramosian-glider, line 104: This branch is unneeded now, right?

melver: Not quite, we can still get an address that is the first (reserved) page that'll pass is_kfence_addr. In this case, the index will be -1.

I added a test for invalid accesses which exercises this.

mm/kfence/core.c
ramosian-glider: We could probably refresh the protections in the heartbeat callback, depending on some dirty flag.

melver: Is it worth it? We already reported one UAF, do we want to report another on the (likely) same address? I don't think we're loosing anything in terms of telling the user there's a bug.

It would also mean iterating through all free objects, which isn't free either, and we might block __kfence_allocs because we need to take the lock of each kfence_metadata.

mm/kfence/kfence-test.c
ramosian-glider: Cool, thanks!

mm/slub.c
ramosian-glider: Yes, I'll try to.

mm/kfence/kfence-test.c
```diff
@@ -422,6 +422,45 @@ static void test_memcache_ctor(struct kunit *test)
...
+	test_free(buf1);
+
+	/* Try to get same address again -- this can take a while. */
+	for (i = 0;; i++) {
+		buf2 = test_alloc(test, size, GFP_KERNEL | __GFP_ZERO, ALLOCATE_ANY);
+		if (buf1 == buf2)
+			break;
+		test_free(buf2);
+
+		if (i == CONFIG_KFENCE_NUM_OBJECTS) {
```
ramosian-glider, line 498: Hitting KFENCE_NUM_OBJECTS doesn't necessarily mean we cannot get the same object again.
But at least it's a reasonable stopping condition.

mm/slab.c
```diff
@@ -3231,7 +3231,7 @@ slab_alloc_node(struct kmem_cache *cachep, gfp_t flags, int nodeid, size_t orig_
 
 	ptr = kfence_alloc(cachep, orig_size, flags);
 	if (ptr)
-		goto leave;
+		goto out_hooks;
```
ramosian-glider, line 3234: But `out_hooks` and `out` themselves are inconsistent?

mm/kfence/core.c
```diff
@@ -251,9 +251,15 @@ static void *kfence_guarded_alloc(struct kmem_cache *cache, size_t size, gfp_t g
...
+	/* Memory initialization. */
+
+	/*
+	 * We check slab_want_init_on_alloc() outselves, rather than letting
+	 * SL*B to the initialization, as otherwise we might overwrite KFENCE's
+	 * redzone.
+	 */
 	addr = (void *)meta->addr;
-	if (gfp & __GFP_ZERO)
+	if (unlikely(slab_want_init_on_alloc(gfp, cache)))
```
ramosian-glider, line 263: Thanks!
I was just going to open a bug for this.

mm/kfence/core.c
```diff
@@ -251,9 +251,15 @@ static void *kfence_guarded_alloc(struct kmem_cache *cache, size_t size, gfp_t g
 
 	raw_spin_unlock_irqrestore(&meta->lock, flags);
 
-	/* Initialization. */
+	/* Memory initialization. */
+
+	/*
+	 * We check slab_want_init_on_alloc() outselves, rather than letting
+	 * SL*B to the initialization, as otherwise we might overwrite KFENCE's
```
ramosian-glider, line 259: s/to/do

mm/kfence/kfence-test.c
melver: It's an approximation. It might have also been allocated by system elsewhere and we're never getting it back. But at least this way we have cycled through the freelist at least once.

mm/kfence/core.c
```diff
@@ -572,6 +572,14 @@ void __kfence_free(void *addr)
 	/* Check canary bytes for memory corruption. */
 	for_each_canary(meta, check_canary_byte);
 
+	/*
+	 * Clear memory if init-on-free is set. While we protect the page, the
+	 * data is still there, and after a use-after-free is detected, we
+	 * unprotect the page, so the data is still accessible.
+	 */
+	if (unlikely(slab_want_init_on_free(meta->cache)))
```
ramosian-glider: Good catch!

mm/slab.c
melver: There is already an 'out' in this function which we cannot jump to.

mm/kfence/kfence-test.c
```diff
@@ -425,6 +425,32 @@ static void test_free_bulk(struct kunit *test)
...
+	/* Assume it hasn't been disabled on command line. */
+
+	setup_test_cache(test, size, NULL);
+	expect.addr = test_alloc(test, size, GFP_KERNEL, ALLOCATE_ANY);
+	for (i = 0; i < size; i++)
+		expect.addr[i] = i + 1;
+	test_free(expect.addr);
+
+	for (i = 0; i < size; i++)
+		KUNIT_EXPECT_EQ(test, expect.addr[i], (char)0);
```
ramosian-glider, line 450: While this is currently true, I don't think it's safe to assume that we'll keep this page unprotected forever.

mm/slab.c
ramosian-glider: Ah, haven't noticed it.

mm/kfence/kfence-test.c
ramosian-glider: Ok.

mm/kfence/core.c
melver: Thanks, I'll push this fix to 'kfence' directly.

mm/slub.c
melver: I still don't know what the right thing here is -- I believe we definitely end up in slab_free_freelist_hook(), and the check is required. But clearly it won't work for us.

Also, we definitely need some of things inside it, e.g. 'slab_free_hook'.

mm/kfence/kfence-test.c
melver: Then we'll just unprotect it again. (The test is checking that we encountered an UAF.) The data must always be accessible and the page protection does not prevent accesses to the data. 

The only way this will fail if the page was reallocated and somebody else wrote to it, however the likelihood of this page having been reallocated to some other KFENCE alloc is near impossible.

I'll try to add a comment.

melver: Oh no I think I know what you mean. I'll fix the test.

PR #111, opened 2020-07-28
kfence: Fix get_stack_skipnr() with tail call optimization

melver:
Test was unhappy after rebase. Seems something caused compiler to optimize more.

Signed-off-by: Marco Elver <elver@google.com>

### DIFF
--- a/mm/kfence/report.c
+++ b/mm/kfence/report.c
@@ -28,7 +28,7 @@ static int get_stack_skipnr(const unsigned long stack_entries[], int num_entries
 			    enum kfence_error_type type)
 {
 	char buf[64];
-	int skipnr;
+	int skipnr, fallback = 0;
 
 	for (skipnr = 0; skipnr < num_entries; skipnr++) {
 		int len = scnprintf(buf, sizeof(buf), "%ps", (void *)stack_entries[skipnr]);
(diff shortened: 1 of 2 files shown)


mm/kfence/report.c
```diff
@@ -43,13 +43,19 @@ static int get_stack_skipnr(const unsigned long stack_entries[], int num_entries
 			break;
 		case KFENCE_ERROR_CORRUPTION:
 		case KFENCE_ERROR_INVALID_FREE:
+			if (!strncmp(buf, "kfence_", sizeof("kfence_") - 1) ||
```
ramosian-glider, line 46: Time to factor away `strncmp(buf, string, strlen(string))`, maybe?

melver: The size is known at compile time, so equivalent would only be a macro. Safer is a strlen, but I don't think it buys us anything and will only add inconsistencies.

If we're not actually saving anything in terms of LOC and the expression is not complex, I usually avoid new indirections as it adds inconsistencies with other subsystems, which undoubtedly could benefit from something like this as well.

In this case, such a primitive should be provided by the kernel somewhere, but they don't, and I think it will invite criticism if we add utilities if we're not even saving LOCs.

melver: Also, this primitive would be called "hasprefix(" or so, and is obviously common in other languages or libraries.

I had the same problem in KCSAN, but decided rolling low-level primitives like this is a bad signal, because the kernel community obviously decided it's not worth it.

melver: Ok, thankfully I was wrong. 

They has `str_has_prefix` and `strstarts`. Now, what is the difference. I'll update the commit to use one of them.

melver: str_has_prefix() it is.

Issue #108, opened 2020-07-27
[kfence] Test with aggressive test config


melver:
Test with the following config. It enables various debugging tools, and enables KFENCE + fault-injection (due to fault-injection, ignore all KFENCE reports).

Currently this config is unable to boot the kernel. Disabling KFENCE results in a bootable kernel again.

There are various issues:
- kmemleak complains
- lockdep complains
- panics eventually

Also try things like (if able to boot):
```
$> echo function > /sys/kernel/debug/tracing/current_tracer
```

[config-kfence.txt](https://github.com/google/kasan/files/4984086/config-kfence.txt)

ramosian-glider:
With `kfence.sample_rate=0` the following report still reproduces:

```
[    4.008908][    T1] x86/mm: Checking user space page tables^M
[    4.009643][    T1] ------------[ cut here ]------------^M
[    4.010218][    T1] x86/mm: Found insecure W+X mapping at address startup_64+0x0/0x30^M
```
(i.e. is unrelated to KFENCE)

melver:
> With `kfence.sample_rate=0` the following report still reproduces:
> 
> ```
> [    4.008908][    T1] x86/mm: Checking user space page tables^M
> [    4.009643][    T1] ------------[ cut here ]------------^M
> [    4.010218][    T1] x86/mm: Found insecure W+X mapping at address startup_64+0x0/0x30^M
> ```
> 
> (i.e. is unrelated to KFENCE)

Ok good, let's see if this one's fixed in newer kernels, although the remaining ones are probably real then. Getting it to boot would be the first step, and then trying to get all the other issues, including kmemleak, to work would be good.

melver:
I think all issues have been fixed now. We can leave it open to continue including this config in tests.

PR #112, opened 2020-07-28
Some fixes

ramosian-glider:
With these patches https://github.com/google/kasan/issues/108 boots to sshd, but freezes soon after that.

### DIFF
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -23,7 +23,7 @@ extern char __kfence_pool[KFENCE_POOL_SIZE];
 
 extern struct static_key_false kfence_allocation_key;
 
-static __always_inline bool is_kfence_addr(void *addr)
+static __always_inline bool is_kfence_addr(const void *addr)
 {
 	return unlikely((char *)addr >= __kfence_pool &&
 			(char *)addr < __kfence_pool + KFENCE_POOL_SIZE);
@@ -65,7 +65,7 @@ bool kfence_handle_page_fault(unsigned long addr);
 // TODO: remove for v1
 // clang-format off
 
-static inline bool is_kfence_addr(void *addr) { return false; }
+static inline bool is_kfence_addr(const void *addr) { return false; }
 static inline void kfence_init(void) { }
 static inline bool kfence_discard_slab(struct kmem_cache *s, struct page *page) { return false; }
 static inline bool kfence_shutdown_cache(struct kmem_cache *s) { return true; }
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -46,7 +46,7 @@ static __always_inline void *kfence_alloc(struct kmem_cache *s, size_t size, gfp
 
 size_t kfence_ksize(const void *addr);
 
-void *kfence_obj_start(const void *addr);
+void *kfence_object_start(const void *addr);
 
 void __kfence_free(void *addr);
 
@@ -71,7 +71,7 @@ static inline bool kfence_discard_slab(struct kmem_cache *s, struct page *page) 
 static inline bool kfence_shutdown_cache(struct kmem_cache *s) { return true; }
 static inline void *kfence_alloc(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
 static inline size_t kfence_ksize(const void *addr) { return 0; }
-static inline void *kfence_obj_start(const void *addr) { return NULL; }
+static inline void *kfence_object_start(const void *addr) { return NULL; }
 static inline bool kfence_free(void *addr) { return false; }
 static inline bool kfence_handle_page_fault(unsigned long addr) { return false; }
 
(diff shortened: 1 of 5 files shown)


mm/kmemleak.c
```diff
@@ -1944,8 +1945,18 @@ void __init kmemleak_init(void)
 	/* register the data/bss sections */
 	create_object((unsigned long)_sdata, _edata - _sdata,
 		      KMEMLEAK_GREY, GFP_ATOMIC);
+#ifndef CONFIG_KFENCE
 	create_object((unsigned long)__bss_start, __bss_stop - __bss_start,
 		      KMEMLEAK_GREY, GFP_ATOMIC);
+#else
+	/* KFENCE objects are located in .bss, which may confuse kmemleak. Skip them. */
```
melver, line 1952: Great! I'm glad we could avoid an is_kfence_addr() this way.

include/linux/kfence.h
```diff
@@ -69,6 +71,7 @@ static inline bool kfence_discard_slab(struct kmem_cache *s, struct page *page)
 static inline bool kfence_shutdown_cache(struct kmem_cache *s) { return true; }
 static inline void *kfence_alloc(struct kmem_cache *s, size_t size, gfp_t flags) { return NULL; }
 static inline size_t kfence_ksize(const void *addr) { return 0; }
+static inline void *kfence_obj_start(const void *addr) { return NULL; }
```
melver, line 74: I think for the main APIs like this, we shouldn't abbreviate.

I could offer: `kfence_object_start()`, `kfence_object_of()`, or `kfence_start_of()`.

I was trying to find symmetry / consistency with some existing names, and it reminds me of `container_of()`.

mm/slub.c
```diff
@@ -3991,6 +3991,7 @@ void __check_heap_object(const void *ptr, unsigned long n, struct page *page,
 	struct kmem_cache *s;
```
melver, line 3991: Isn't this function also in slab.c, and is it also broken?

mm/kmemleak.c
```diff
@@ -96,6 +96,7 @@
 #include <asm/processor.h>
 #include <linux/atomic.h>
 
+#include <linux/kfence.h>
```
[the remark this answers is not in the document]
ramosian-glider: Yes, this was my intent, but I failed at sorting "kasan" and "kfence".

mm/slub.c
ramosian-glider: Correct, will fix that separately.

mm/kfence/core.c
```diff
@@ -276,7 +276,6 @@ static void *kfence_guarded_alloc(struct kmem_cache *cache, size_t size, gfp_t g
 		/* Randomly inject "faults" by protecting the allocated object. */
 		kfence_protect(meta->addr);
 	}
-
```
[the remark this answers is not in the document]
ramosian-glider: No, I accidentally deleted it when adding/deleting pr_debug calls. Fixed.
Also removed the braces around kfence_protect() above.

include/linux/kfence.h
ramosian-glider: I think kfence_object_start() is more descriptive.

mm/kfence/core.c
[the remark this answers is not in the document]
melver: I recently had this comment: https://lkml.kernel.org/r/20200721140929.GB10769@hirez.programming.kicks-ass.net -- after which I did add {} where technically it wasn't needed.

Peter has a point: if, whatever comes after the 'if (' isn't just on 1 line (or 2, if generous), add "bucket of braces". Otherwise it might hurt readability.

I leave it to you. I'm indifferent, but am trying to refine the style to follow kernel's best-practices.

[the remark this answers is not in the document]
ramosian-glider: I think moving the comment before the if-statement (which I also did) solves this problem.
I agree that having more than one line without the braces may worsen the readability.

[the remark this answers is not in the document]
melver: Although the comment only applies if the config option is set. We could shorten the comment to

```
/* Random "faults" by protecting the object. */
```

And place it on the same line as kfence_protect().

Sorry for the bike-shedding. ;-)

mm/slub.c
ramosian-glider: Done.

ramosian-glider:
Added a couple more commits: const for is_kfence_addr() and hardened usercopy support for SLAB.

include/linux/kfence.h
```diff
@@ -23,7 +23,7 @@ extern char __kfence_pool[KFENCE_POOL_SIZE];
 
 extern struct static_key_false kfence_allocation_key;
 
-static __always_inline bool is_kfence_addr(void *addr)
+static __always_inline bool is_kfence_addr(const void *addr)
```
melver, line 26: Do we eventually have to add volatile as well? Because we will get the inevitable warnings if somebody passes a volatile pointer.

melver: Hmm, k{free,size}() only take 'const void *'. So maybe we can ignore it for now.

PR #115, opened 2020-07-30
Assorted commits

melver:
The major fix in here being a fix for the lockups with the config in https://github.com/google/kasan/issues/108.

### DIFF
--- a/arch/x86/include/asm/kfence.h
+++ b/arch/x86/include/asm/kfence.h
@@ -137,7 +137,7 @@ static bool arch_kfence_initialize_pool(void)
 {
 	unsigned long addr = (unsigned long)__kfence_pool;
 
-	while (is_kfence_addr((void *)addr)) {
+	while (is_kfence_address((void *)addr)) {
 		unsigned int level;
 		pte_t *pte = lookup_address(addr, &level);
 
--- a/arch/x86/include/asm/kfence.h
+++ b/arch/x86/include/asm/kfence.h
@@ -13,6 +13,12 @@
 #define KFENCE_POOL_ALIGNMENT PAGE_SIZE
 
 /*
+ * The page fault handler entry function, up to which the stack trace is
+ * truncated in reports.
+ */
+#define KFENCE_SKIP_ARCH_FAULT_HANDLER "asm_exc_page_fault"
+
+/*
  * TODO: Clean this up.
  */
 
@@ -133,7 +139,7 @@ static bool split_large_page(pte_t *kpte, unsigned long address, unsigned int le
 	return true;
 }
 
-static bool arch_kfence_initialize_pool(void)
+static inline bool arch_kfence_initialize_pool(void)
 {
 	unsigned long addr = (unsigned long)__kfence_pool;
 
@@ -154,7 +160,7 @@ static bool arch_kfence_initialize_pool(void)
 	return true;
 }
 
-static bool kfence_change_page_prot(unsigned long addr, bool protect)
+static inline bool kfence_change_page_prot(unsigned long addr, bool protect)
 {
 	unsigned int level;
 	pte_t *pte = lookup_address(addr, &level);
--- a/mm/kfence/kfence.h
+++ b/mm/kfence/kfence.h
@@ -32,7 +32,7 @@ enum kfence_object_state {
 
 /* KFENCE metadata per guarded allocation. */
 struct kfence_metadata {
-	struct list_head list; /* Freelist node. */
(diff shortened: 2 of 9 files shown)


melver:
https://github.com/google/kasan/issues/108 seems to be booting + passing the test (modulo some flakiness because it seems kmemleak hammers kmalloc() as well and a lot of KFENCE allocations go to kmemleak).

melver:
I'll try to refrain from pushing more commits now until this set it merged.

[ Currently working on cleaning up x86's <asm/kfence.h>, but that can wait. ]

mm/slub.c
```diff
@@ -1484,12 +1484,10 @@ static inline bool slab_free_freelist_hook(struct kmem_cache *s,
 	void *old_tail = *tail ? *tail : *head;
 	int rsize;
 
-	/*
-	 * TODO: slab_free_freelist_hook() doesn't work with KFENCE-allocated
-	 * addresses. Need to handle them separately.
-	 */
-	if (is_kfence_address(*head))
+	if (is_kfence_address(next)) {
+		slab_free_hook(s, next);
```
melver, line 1488: I haven't tested with SLAB yet, but I assume it has the same problem.

mm/kfence/report.c
```diff
@@ -132,6 +132,15 @@ void kfence_report_error(unsigned long address, const struct kfence_metadata *me
 	int num_stack_entries = stack_trace_save(stack_entries, KFENCE_STACK_DEPTH, 1);
 	int skipnr = get_stack_skipnr(stack_entries, num_stack_entries, type);
 
+	lockdep_assert_held(&meta->lock);
+	/*
+	 * Disable lockdep, as it might report on potential deadlock depending
+	 * on where printk() is called; this is unfortunately expected and
```
ramosian-glider, line 138: The "expected and unavoidable" part is still lacking some information. Is this report expected because it's a real deadlock? If it is unavoidable, could it break us?

melver: Yes, it could break us, if a report is generated from the wrong place in the kernel. But this problem exists in all debugging tools (in fact when I had kmemleak enabled, I got lockdep warnings about kmemleak when it started printing things).

Calling printk from certain parts of the kernel may lead to deadlock, e.g. when called from the scheduler. I think that's what lockdep has been warning about. I had the same problem in KCSAN, and all we can do is (other than disable the tool in those parts of the kernel), is disable lockdep and ensure that those bits of the kernel don't have real bugs.

It's a known problem: https://lore.kernel.org/linuxppc-dev/20200623202404.GE2483@worktop.programming.kicks-ass.net/

I'll try to add comment that explains it better.

mm/kfence/kfence.h
```diff
@@ -17,8 +17,8 @@
 #define PTR_FMT "%p"
 #endif
 
-/* Helper to get the canary byte pattern for @addr. */
-#define KFENCE_CANARY_PATTERN(addr) (((u8[]){ 0xaa, 0xab, 0xaa, 0xad })[(size_t)addr % 4])
+/* Get the canary byte pattern for @addr. */
+#define KFENCE_CANARY_PATTERN(addr) ((u8)0xaa ^ (u8)((unsigned long)addr & 0x7))
```
ramosian-glider, line 21: Let's add some comment describing why we're using varying bytes.
E.g. "Use a rolling 8-byte pattern to detect memory corruptions with higher probability."

mm/slub.c
ramosian-glider: I've been looking into removing this check altogether.

ramosian-glider: But this is also fine.

arch/x86/include/asm/kfence.h
```diff
@@ -12,6 +12,12 @@
 /* TODO: Decide if we want huge page or 4k page on x86. */
 #define KFENCE_POOL_ALIGNMENT PAGE_SIZE
 
+/*
+ * The page fault handler entry function, up to which the stack trace is
+ * truncated in reports.
+ */
+#define KFENCE_SKIP_ARCH_FAULT_HANDLER "asm_exc_page_fault"
```
ramosian-glider, line 19: Maybe checking for "page_fault" was a bearable approximation?

melver: There are multiple functions with "page_fault" in the name in the stack trace, and therefore doesn't work. Also, for ARM it's going to be different again.

mm/kfence/kfence.h
melver: Done.

PR #116, opened 2020-07-31
Simplify x86 page splitting

melver:


melver:
> Thanks!
> It's strange that we cannot reproduce the conditions under which the split is required, but perhaps upstream reviewers will help us with that.

I think it's still required, and as far as I can tell, set_memory_4k() is executed exactly once with the default config. That makes sense, since the default uses 2M pool, and that is exactly HPAGE_SIZE.

PR #117, opened 2020-07-31
Allow using KFENCE together with KASAN

ramosian-glider:


mm/kasan/common.c
```diff
@@ -493,6 +500,7 @@ static void *__kasan_kmalloc(struct kmem_cache *cache, const void *object,
 	if (cache->flags & SLAB_KASAN)
 		set_track(&get_alloc_info(cache, object)->alloc_track, flags);
 
+out:
 	return set_tag(object, tag);
```
melver, line 504: What happens if we set a tag?

ramosian-glider: Hm, actually, we'd better skip this path. It doesn't matter for KASAN, but KHWASAN is a different story.

ramosian-glider: Removed the goto

include/linux/slub_def.h
```diff
@@ -7,6 +7,7 @@
  *
  * (C) 2007 SGI, Christoph Lameter
  */
+#include <linux/kasan.h>
```
melver, line 10: Should we send patches to fix KFENCE unrelated things before RFC?

If we can get this fixed in mainline before would be good, as it saves us a patch in the series. WDYT?

ramosian-glider:
Sure, I'll send this patch separately.

melver:
> Sure, I'll send this patch separately.

Hmm, I investigated a bit more and for some reason include/linux/sl*b_def.h don't include dependencies for structs on purpose it seems. So my guess is, such a patch will be rejected.

We just have to include <linux/slab.h> in kfence/kfence.h, which should solve it (because linux/slab.h includes kasan.h and hopefully any other future dependencies? although it is a bit of a mess..). I have some more fixes queued and will send Monday.

PR #118, opened 2020-08-03
Minor fixes

melver:
Not quite sure what the best strategy for handling the sl*b_def.h mess is. We might be better off not touching those files. WDYT?

### DIFF
--- a/lib/Kconfig.kfence
+++ b/lib/Kconfig.kfence
@@ -5,6 +5,7 @@ menuconfig KFENCE
 	bool "KFENCE: low-overhead use-after-free and buffer-overflow detector"
 	depends on HAVE_ARCH_KFENCE && (!KASAN || EXPERT) && (SLAB || SLUB)
 	depends on JUMP_LABEL # To ensure performance, require jump labels
+	select STACKTRACE
 	help
 	  KFENCE is low-overhead sampling-based detector for heap
 	  use-after-free and buffer-overflow errors. KFENCE is designed to have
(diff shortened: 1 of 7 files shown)


mm/kfence/report.c
```diff
@@ -131,7 +131,8 @@ void kfence_report_error(unsigned long address, const struct kfence_metadata *me
 	int num_stack_entries = stack_trace_save(stack_entries, KFENCE_STACK_DEPTH, 1);
 	int skipnr = get_stack_skipnr(stack_entries, num_stack_entries, type);
 
-	lockdep_assert_held(&meta->lock);
+	if (meta)
```
ramosian-glider, line 135: This isn't the only place dereferencing `meta` in this function, shall we do anything about the rest?

lib/Kconfig.kfence
```diff
@@ -5,6 +5,7 @@ menuconfig KFENCE
 	bool "KFENCE: low-overhead use-after-free and buffer-overflow detector"
 	depends on HAVE_ARCH_KFENCE && (!KASAN || EXPERT) && (SLAB || SLUB)
 	depends on JUMP_LABEL # To ensure performance, require jump labels
+	select STACKTRACE
```
ramosian-glider, line 8: Didn't know one could unselect it.

melver: Yeah, make tinyconfig did not select it. (Some eventual random config would have hit this and we'd have gotten test-robot complaints.)

mm/kfence/report.c
melver: Only for OOB it's non-optional. I added a precondition check.

PR #123, opened 2020-08-04
Support ARM64

melver:
Add support for ARM64.

It appears that the static pool cannot be used on ARM64 trivially, because it doesn't set up valid pages for statically allocated memory (at least `virt_to_page()` always returned an invalid page pointer). At some future point, we may want to try and understand if we can force it to give us some pages.

I will try to research more what exactly happens with .bss memory on ARM64 for the commit message. If you know what's happening, please let me know.

All tests pass in a VM (except the test_gfpzero is a bit flaky due to it not getting the same object back)!

(I'll wait to push this until your API documentation changes are in the repo, after which I'll do a rebase, as there are some conflicts.)

### DIFF
--- a/arch/x86/Kconfig
+++ b/arch/x86/Kconfig
@@ -333,6 +333,10 @@ config KASAN_SHADOW_OFFSET
 	depends on KASAN
 	default 0xdffffc0000000000
 
+config KFENCE_STATIC_POOL
+	def_bool y
+	depends on KFENCE
+
 config HAVE_INTEL_TXT
 	def_bool y
 	depends on INTEL_IOMMU && ACPI
--- a/arch/x86/Kconfig
+++ b/arch/x86/Kconfig
@@ -144,6 +144,7 @@ config X86
 	select HAVE_ARCH_KASAN			if X86_64
 	select HAVE_ARCH_KASAN_VMALLOC		if X86_64
 	select HAVE_ARCH_KFENCE
+	select HAVE_ARCH_KFENCE_STATIC_POOL
 	select HAVE_ARCH_KGDB
 	select HAVE_ARCH_MMAP_RND_BITS		if MMU
 	select HAVE_ARCH_MMAP_RND_COMPAT_BITS	if MMU && COMPAT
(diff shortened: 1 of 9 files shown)


melver:
> I think having loads at malloc/free fast paths is basically a no-go for Android.
> We can look into allocating the pool at a static address, same as KASAN does.

It's only the free paths. Also, do notice that we call kfence_free() in a free slow-path (per `stat(s, FREE_SLOWPATH)` above the `kfence_free()`).

I can add a big `TODO` in ARM64's Kconfig, so we can resolve this. I'd still like to merge this if there is nothing else wrong with it, because it at least gives us something to play with. We can even benchmark it and see what the performance situation is on a real device.

ramosian-glider:
Fair enough. Let me take another look at the patches.

arch/x86/Kconfig
```diff
@@ -333,6 +333,10 @@ config KASAN_SHADOW_OFFSET
 	depends on KASAN
 	default 0xdffffc0000000000
 
+config KFENCE_STATIC_POOL
```
ramosian-glider, line 336: Shouldn't it be something like HAVE_KFENCE_STATIC_POOL that is always true on x86?
Or do you want a possibility to test the non-static pool as well?

melver: This is an immutable Kconfig option, although going with some HAVE_ option is probably better. I'll change it.

melver:
> > I think having loads at malloc/free fast paths is basically a no-go for Android.
> > We can look into allocating the pool at a static address, same as KASAN does.
> 
> It's only the free paths. Also, do notice that we call kfence_free() in a free slow-path (per `stat(s, FREE_SLOWPATH)` above the `kfence_free()`).
> 
> I can add a big `TODO` in ARM64's Kconfig, so we can resolve this. I'd still like to merge this if there is nothing else wrong with it, because it at least gives us something to play with. We can even benchmark it and see what the performance situation is on a real device.

One more thing: I noticed in the old implementation that the pool pointer was likely on the same cacheline as the spinlock (they were right next to each other) which certainly is unfavorable due to cache-line ping-pong. This new version uses __read_mostly for the pool pointer which avoids this. So, I do suggest re-benchmarking this version on a real ARM device if it turns out using a static pool somehow doesn't work due to inherent limitations of how pages are set up for .bss data (hopefully not).

PR #124, opened 2020-08-04
added LICENSE to project

jivthesh:


melver:
There is no explanation, or justification. Can you explain why?

The master branch only contains a README summarizing KASAN, and the code itself is all in the Linux kernel. Furthermore, KASAN itself is GPL since it lives in the Linux kernel, so this LICENSE doesn't even make sense.

PR #128, opened 2020-08-05
Bulk allocations

ramosian-glider:


mm/kfence/kfence-test.c
```diff
@@ -625,15 +625,51 @@ static void test_krealloc(struct kunit *test)
...
  */
 // clang-format off
 // TODO: fix formatting for v1
-#define KFENCE_KUNIT_CASE(test_name)                                           \
-	{ .run_case = test_name, .name = #test_name },                         \
+#define KFENCE_KUNIT_CASE(test_name)						\
+	{ .run_case = test_name, .name = #test_name },				\
+	{ .run_case = test_name, .name = #test_name "-memcache" }
+
+#define KFENCE_MEMCACHE_KUNIT_CASE(test_name)					\
```
melver, line 672: This one is not needed. The "trick" is to add memcache into the test name, as is done for some other tests. So your test could simply be named "test_memcache_alloc_bulk", which is also more descriptive.

I should document this somewhere.

ramosian-glider: Ah, got it. I've noticed the comparison against "memcache", but didn't realized it is normally applied to function names.

mm/kfence/kfence-test.c
```diff
@@ -625,15 +625,51 @@ static void test_krealloc(struct kunit *test)
...
+	 */
+	timeout = jiffies + msecs_to_jiffies(100 * CONFIG_KFENCE_SAMPLE_INTERVAL);
+	do {
+		num = kmem_cache_alloc_bulk(test_cache, GFP_ATOMIC, ARRAY_SIZE(objects), objects);
+		if (!num)
+			continue;
+		for (i = 0; i < ARRAY_SIZE(objects); i++) {
+			if (is_kfence_address(objects[i])) {
+				pass = true;
+				break;
```
melver, line 651: Doesn't "!pass" take care of this? It seems we're leaking memory.

mm/kfence/kfence-test.c
```diff
@@ -625,15 +625,51 @@ static void test_krealloc(struct kunit *test)
 	KUNIT_ASSERT_TRUE(test, report_matches(&expect));
 }
 
+/* Test that some objects from a bulk allocation belong to KFENCE pool. */
+static void test_alloc_bulk(struct kunit *test)
+{
+	const size_t size = 32;
+	void *objects[100];
+	bool pass = false;
+	int i, num;
```
melver, line 634: Move 'i,num' into the do-loop, and just initialize num straight away?

mm/kfence/kfence-test.c
```diff
@@ -625,15 +625,51 @@ static void test_krealloc(struct kunit *test)
 	KUNIT_ASSERT_TRUE(test, report_matches(&expect));
 }
 
+/* Test that some objects from a bulk allocation belong to KFENCE pool. */
+static void test_alloc_bulk(struct kunit *test)
+{
+	const size_t size = 32;
+	void *objects[100];
```
melver, line 632: Can also move into the do-loop block?

mm/kfence/kfence-test.c
melver: My bad, I missed the nested loop.

mm/kfence/kfence-test.c
ramosian-glider: Done

mm/kfence/kfence-test.c
ramosian-glider: Done

mm/slab.c
```diff
@@ -3529,7 +3529,10 @@ int kmem_cache_alloc_bulk(struct kmem_cache *s, gfp_t flags, size_t size,
 
 	local_irq_disable();
 	for (i = 0; i < size; i++) {
-		void *objp = __do_cache_alloc(s, flags);
+		void *objp = kfence_alloc(s, s->object_size, flags);
+
+		if (likely(!objp))
```
melver, line 3534: In the kernel we may use the GNU C extension to the conditional operator (used all over the place): https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html

So this could be
```
void *objp = kfence_alloc(s, s->object_size, flags) ?: __do_cache_alloc(s, flags);
```

I leave it to you if that's more readable or not.

mm/kfence/kfence-test.c
```diff
@@ -625,15 +625,55 @@ static void test_krealloc(struct kunit *test)
...
+	do {
+		int i, num;
+		void *objects[100];
+
+		/*
+		 * kmem_cache_alloc_bulk() disables interrupts, and calling it
+		 * in a tight loop may not give KFENCE a chance to switch the
+		 * static branch. Call cond_resched() to let KFENCE chime in.
+		 */
+		cond_resched();
```
melver, line 651: We can move this to the end of the loop, as the resched in the first iteration is probably unnecessary as we haven't tried to allocate anything yet.

Then you can also initialize num straight away.

mm/kfence/kfence-test.c
```diff
@@ -625,15 +625,55 @@ static void test_krealloc(struct kunit *test)
...
+		num = kmem_cache_alloc_bulk(test_cache, GFP_ATOMIC, ARRAY_SIZE(objects), objects);
+		if (!num)
+			continue;
+		for (i = 0; i < ARRAY_SIZE(objects); i++) {
+			if (is_kfence_address(objects[i])) {
```
melver, line 656: Should we check that we have both kfence and non-kfence objects?

mm/kfence/kfence-test.c
ramosian-glider: Done.

mm/kfence/kfence-test.c
ramosian-glider: I don't think that's necessary. If for some reason we always allocate KFENCE objects, we would be out of them by this point already.

mm/slab.c
ramosian-glider: Hm, that's handy and probably more readable.

PR #132, opened 2020-08-06
Documentation

melver:


Documentation/dev-tools/kfence.rst
```diff
@@ -17,33 +17,185 @@ uptime is when the tool is deployed across a large fleet of machines.
...
+
+The most important parameter is KFENCE's sample interval, which can be set via
+the kernel boot parameter `kfence.sample_interval` in milliseconds. The sample
+interval determines the frequency with which heap allocations will be guarded
+by KFENCE. The default is configurable via the Kconfig option
+``CONFIG_KFENCE_SAMPLE_INTERVAL``.
+
+With the Kconfig option ``CONFIG_KFENCE_NUM_OBJECTS`` (default 255), the number
+of available guarded objects can be controlled. Each object requires 2 pages,
+with one being the object page, and the 2 adjacent pages used as guard pages.
```
ramosian-glider, line 38: This reads as if every object required 3 pages.
How about something along the lines of:

> Each object requires 2 pages, one for the object itself and the other one used as a guard page.
> Object pages are interleaved with guard pages, so every object page is surrounded by two guard pages.

melver:
> Do we want to mention the rare "REDZONE : OBJECT: REDZONE" case somewhere?

Added in brackets.

PR #133, opened 2020-08-07
kfence: fix kfence_debugfs_init() declaration, should be returning int

ramosian-glider:


ramosian-glider:
Spotted when porting to Android.

PR #134, opened 2020-08-17
Lockdep compatibility

melver:
Lockdep complains if certain structures are in static data. Exclude kfence memory.

If this problem keeps coming up again in other contexts, we might need to look for a better solution.

### DIFF
--- a/kernel/locking/lockdep.c
+++ b/kernel/locking/lockdep.c
@@ -38,6 +38,7 @@
 #include <linux/seq_file.h>
 #include <linux/spinlock.h>
 #include <linux/kallsyms.h>
+#include <linux/kfence.h>
 #include <linux/interrupt.h>
 #include <linux/stacktrace.h>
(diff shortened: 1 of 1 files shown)


kernel/locking/lockdep.c
```diff
@@ -755,6 +756,9 @@ static int static_obj(const void *obj)
 	if (arch_is_kernel_initmem_freed(addr))
 		return 0;
 
+	if (is_kfence_address(obj))
```
ramosian-glider, line 759: Can you please add a comment here?
Something along the lines of "Lockdep complains if certain structures are in static data. Exclude kfence memory." should go.

melver: The function 'static_obj' only checks if something is a static object. I added the appropriate comment.

ramosian-glider:
LGTM, thanks!

PR #135, opened 2020-09-07
kfence: Remove clang-format on/off

melver:
This is to get the rebased patch-series ready. From this point on, we should probably use clang-format-diff.

I could probably just apply this to the kfence-rebase-wip branch, but it'll make cherry-picking onto + rebasing that branch more difficult.


ramosian-glider (approved):
Thanks!

PR #137, opened 2020-09-08
kfence: kmemleak: use delete_object_part() for KFENCE pool

ramosian-glider:
As suggested by Catalin Marinas


melver:
Thanks!

PR #138, opened 2020-09-08
kfence: Reporting and documentation fixes for v2

melver:
Andrey, let me know if this clarifies the Documentation.

Thanks,
-- Marco

### DIFF
--- a/Documentation/dev-tools/kfence.rst
+++ b/Documentation/dev-tools/kfence.rst
@@ -40,11 +40,12 @@ guarded by KFENCE. The default is configurable via the Kconfig option
 ``CONFIG_KFENCE_SAMPLE_INTERVAL``. Setting ``kfence.sample_interval=0``
 disables KFENCE.
 
-With the Kconfig option ``CONFIG_KFENCE_NUM_OBJECTS`` (default 255), the number
-of available guarded objects can be controlled. Each object requires 2 pages,
-one for the object itself and the other one used as a guard page; object pages
-are interleaved with guard pages, and every object page is therefore surrounded
-by two guard pages.
+The KFENCE memory pool is of fixed size, and if the pool is exhausted, no
+further KFENCE allocations occur. With ``CONFIG_KFENCE_NUM_OBJECTS`` (default
+255), the number of available guarded objects can be controlled. Each object
+requires 2 pages, one for the object itself and the other one used as a guard
+page; object pages are interleaved with guard pages, and every object page is
+therefore surrounded by two guard pages.
 
 The total memory dedicated to the KFENCE memory pool can be computed as::
 
--- a/Documentation/dev-tools/kfence.rst
+++ b/Documentation/dev-tools/kfence.rst
@@ -163,7 +163,7 @@ These are reported on frees::
     ==================================================================
     BUG: KFENCE: memory corruption in test_kmalloc_aligned_oob_write+0xef/0x184
 
-    Detected corrupted memory at 0xffffffffb6797ff9 [ 0xac . . . . . . ] (in kfence-#69):
+    Corrupted memory at 0xffffffffb6797ff9 [ 0xac . . . . . . ] (in kfence-#69):
      test_kmalloc_aligned_oob_write+0xef/0x184
      kunit_try_run_case+0x51/0x85
      kunit_generic_run_threadfn_adapter+0x16/0x30
--- a/Documentation/dev-tools/kfence.rst
+++ b/Documentation/dev-tools/kfence.rst
@@ -21,6 +21,12 @@ To enable KFENCE, configure the kernel with::
 
     CONFIG_KFENCE=y
 
+To build a kernel with KFENCE support, but disabled by default (to enable, set
+``kfence.sample_interval`` to non-zero value), configure the kernel with::
+
+    CONFIG_KFENCE=y
+    CONFIG_KFENCE_SAMPLE_INTERVAL=0
+
 KFENCE provides several other configuration options to customize behaviour (see
 the respective help text in ``lib/Kconfig.kfence`` for more info).
 
@@ -281,9 +287,11 @@ found in the userspace `Electric Fence Malloc Debugger
 In the kernel, several tools exist to debug memory access errors, and in
 particular KASAN can detect all bug classes that KFENCE can detect. While KASAN
 is more precise, relying on compiler instrumentation, this comes at a
-performance cost. We want to highlight that KASAN and KFENCE are complementary,
-with different target environments. For instance, KASAN is the better
-debugging-aid, where a simple reproducer exists: due to the lower chance to
-detect the error, it would require more effort using KFENCE to debug.
-Deployments at scale, however, would benefit from using KFENCE to discover bugs
-due to code paths not exercised by test cases or fuzzers.
+performance cost.
+
+It is worth highlighting that KASAN and KFENCE are complementary, with
+different target environments. For instance, KASAN is the better debugging-aid,
+where test cases or reproducers exists: due to the lower chance to detect the
+error, it would require more effort using KFENCE to debug. Deployments at scale
+that cannot afford to enable KASAN, however, would benefit from using KFENCE to
+discover bugs due to code paths not exercised by test cases or fuzzers.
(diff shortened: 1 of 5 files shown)


Documentation/dev-tools/kfence.rst
```diff
@@ -281,9 +287,11 @@ found in the userspace `Electric Fence Malloc Debugger
...
 is more precise, relying on compiler instrumentation, this comes at a
-performance cost. We want to highlight that KASAN and KFENCE are complementary,
-with different target environments. For instance, KASAN is the better
-debugging-aid, where a simple reproducer exists: due to the lower chance to
-detect the error, it would require more effort using KFENCE to debug.
-Deployments at scale, however, would benefit from using KFENCE to discover bugs
-due to code paths not exercised by test cases or fuzzers.
+performance cost.
+
+We want to highlight that KASAN and KFENCE are complementary, with different
```
ramosian-glider, line 293: Not sure we need to write in first person here, it's only a piece of documentation, not a patch description or an article.
How about: "It is worth highlighting that..."?

Documentation/dev-tools/kfence.rst
```diff
@@ -40,11 +40,12 @@ guarded by KFENCE. The default is configurable via the Kconfig option
 ``CONFIG_KFENCE_SAMPLE_INTERVAL``. Setting ``kfence.sample_interval=0``
 disables KFENCE.
 
-With the Kconfig option ``CONFIG_KFENCE_NUM_OBJECTS`` (default 255), the number
-of available guarded objects can be controlled. Each object requires 2 pages,
-one for the object itself and the other one used as a guard page; object pages
-are interleaved with guard pages, and every object page is therefore surrounded
-by two guard pages.
+The KFENCE memory pool is of fixed size, and if the pool is exhausted, no
```
ramosian-glider, line 43: It just came to my mind that KFENCE adds some implicit memory cost by splitting bigger pages and potentially adding another level to certain page table entries.
Not sure it is worth mentioning though.

ramosian-glider:
LGTM (see one nit above)

Documentation/dev-tools/kfence.rst
melver: It's quite complicated to explain, without going into architecture-specific details. Since it's architecture-specific, we can probably ignore it in Documentation, and if it's important, can just mention it in the various <asm/kfence.h> files.

Documentation/dev-tools/kfence.rst
melver: Done.

Documentation/dev-tools/kfence.rst
```diff
@@ -78,14 +85,15 @@ A typical out-of-bounds access looks like this::
...
+its origin. Note that, real kernel addresses are only shown for
+``CONFIG_DEBUG_KERNEL=y`` builds.
 
 Use-after-free accesses are reported as::
 
     ==================================================================
     BUG: KFENCE: use-after-free in test_use_after_free_read+0xb3/0x143
 
-    Use-after-free access at 0xffffffffb673dfe0:
+    Use-after-free access at 0xffffffffb673dfe0 (in kfence-#24):
```
xairy, line 96: IMO there's no need for parentheses around `kfence-#24`.

Documentation/dev-tools/kfence.rst
```diff
@@ -155,7 +163,7 @@ These are reported on frees::
     ==================================================================
     BUG: KFENCE: memory corruption in test_kmalloc_aligned_oob_write+0xef/0x184
 
-    Detected corrupted memory at 0xffffffffb6797ff9 [ 0xac . . . . . . ]:
+    Detected corrupted memory at 0xffffffffb6797ff9 [ 0xac . . . . . . ] (in kfence-#69):
```
xairy, line 166: Same here about parentheses.

Also, this line and some others in KFENCE reports are quite long. I very much dislike when the lines of some kernel output get trimmed or split across multiple lines. Should we keep all report lines < 80 characters?

Documentation/dev-tools/kfence.rst
```diff
@@ -176,8 +184,13 @@ These are reported on frees::
...
 
-For such errors, the address where the corruption as well as the corrupt bytes
-are shown.
+For such errors, the address where the corruption as well as the invalidly
+written bytes (offset from the address) are shown; in this representation, '.'
+denote untouched bytes. In the example above ``0xac`` is the value written to
+the invalid address at offset 0, and the remaining '.' denote that no following
+bytes have been touched. Note that, real values are only shown for
+``CONFIG_DEBUG_KERNEL=y`` builds; for non-debug builds, '!' is used instead to
+denote invalidly written bytes.
```
xairy, line 193: and `to avoid information disclosure`?

Documentation/dev-tools/kfence.rst
```diff
@@ -211,26 +224,24 @@ Implementation Details
...
-The static branch is toggled to redirect the allocation to KFENCE.
+of the sample interval, the next allocation through the main allocator (SLAB or
+SLUB) returns a guarded allocation from the KFENCE object pool. At this point,
+the timer is reset, and the next allocation is set up after the expiration of
+the interval.  To "gate" a KFENCE allocation through the main allocator's
+fast-path without overhead, KFENCE relies on static branches via the static
+keys infrastructure.  The static branch is toggled to redirect the allocation
+to KFENCE.
 
 KFENCE objects each reside on a dedicated page, at either the left or right
```
xairy, line 235: Are the objects still aligned to `ARCH_SLAB_MINALIGN` when the guard page is on the right side? I'd clarify this here or somewhere.

Documentation/dev-tools/kfence.rst
```diff
@@ -211,26 +224,24 @@ Implementation Details
...
-out-of-bounds access. The side opposite of an object's guard page is used as a
-pattern-based redzone, to detect out-of-bounds writes on the unprotected sed of
-the object on frees (for special alignment and size combinations, both sides of
-the object are redzoned).
+out-of-bounds access.
 
-KFENCE also uses pattern-based redzones on the other side of an object's guard
-page, to detect out-of-bounds writes on the unprotected side of the object;
-these are reported on frees.
+KFENCE also uses pattern-based redzones on the opposite side of an object's
```
xairy, line 242: Still unclear. Logically, a redzone should be on the other side of an object. And the text says that a redzone is on the other side of a guard page. I guess both are true, but these two redzones are guarding two different objects. I'd rather refer to `the other side of an object`.

xairy (changes requested):

Documentation/dev-tools/kfence.rst
melver: In the case here we can do s/corrupted memory/corruption/, which should shorten this a bit. In the case here it'll become 77 chars with the parens removal.

Not sure what other lines are left. Trying to force all lines to be <80 might be hard though, and splitting this text across multiple lines is bad too because it makes grepping etc. impossible. Trying to balance readability, number of characters, and avoiding unnecessary lines requires trade-offs. I'm willing to go over 80 chars if there is no better option.

Documentation/dev-tools/kfence.rst
melver: Not sure how to put this more succinctly, but it's technically: object xor page => redzone. The figure illustrates what we have best.

If this text here somehow mismatches what you infer from the figure, it needs to be fixed until it reads right. :-) I'll try again.

Documentation/dev-tools/kfence.rst
melver: Yes, we honor the slab's requested alignment, and implicitly ARCH_SLAB_MINALIGN (create_cache -> calculate_alignment -> ARCH_SLAB_MINALIGN).

Can add this.

Documentation/dev-tools/kfence.rst
melver: I'm going with s/Detected c/C/ so it'll say "Corrupted memory at ...".

Documentation/dev-tools/kfence.rst
melver: Ok, I just made that change and it reads wrong.

Without parenthesis you'll get this:

"Use-after-free access at 0xffffffffb673dfe0 in kfence-#24:"

which reads wrong, because it wants a ", which is in". Without that, it becomes more ambiguous and you could interpret the "in" not referring to a memory location but as referring to a function.

By using parenthesis, we can get away without ", which is in", since they more strongly associate to the preceding word/value.

Documentation/dev-tools/kfence.rst
melver: Done.

Documentation/dev-tools/kfence.rst
ramosian-glider: I second keeping the parenthesis.

xairy (approved):
LGTM, thanks!

melver:
> LGTM, thanks!

Thank you!

PR #142, opened 2020-09-10
kfence: v2: address comments by Dmitry Vyukov

ramosian-glider:
- do not store negative size for right-located objects
 - only cache-align addresses of right-located objects
 - run toggle_allocation_gate() after KFENCE is enabled

### DIFF
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -212,8 +212,7 @@ static inline void for_each_canary(const struct kfence_metadata *meta, bool (*fn
 			break;
 	}
 
-	for (addr = meta->addr + meta->size; addr < PAGE_ALIGN(meta->addr);
-	     addr++) {
+	for (addr = meta->addr + meta->size; addr < PAGE_ALIGN(meta->addr); addr++) {
 		if (!fn((u8 *)addr))
 			break;
 	}
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -203,7 +203,6 @@ static inline bool check_canary_byte(u8 *addr)
 
 static inline void for_each_canary(const struct kfence_metadata *meta, bool (*fn)(u8 *))
 {
-	const int size = abs(meta->size);
 	unsigned long addr;
 
 	lockdep_assert_held(&meta->lock);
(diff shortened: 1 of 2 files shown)


mm/kfence/core.c
```diff
@@ -632,7 +633,7 @@ size_t kfence_ksize(const void *addr)
 	 * Read locklessly -- if there is a race with __kfence_alloc(), this is
 	 * either a use-after-free or invalid access.
 	 */
-	return meta ? abs(meta->size) : 0;
+	return meta ? meta->size : 0;
```
dvyukov, line 636: Should we also make the size field size_t? We cast it to size_t and sizes are generally passed as size_t in slab code and memory overhead is not a concern for the meta descriptors.

dvyukov:
Otherwise looks good to me.

mm/kfence/core.c
```diff
@@ -213,7 +212,8 @@ static inline void for_each_canary(const struct kfence_metadata *meta, bool (*fn
 			break;
 	}
 
-	for (addr = meta->addr + size; addr < PAGE_ALIGN(meta->addr); addr++) {
+	for (addr = meta->addr + meta->size; addr < PAGE_ALIGN(meta->addr);
+	     addr++) {
```
melver, line 216: This should not line-break (I tried to clang-format it and it doesn't with our .clang-format). Is your clang-format-diff picking up the right .clang-format?

mm/kfence/core.c
melver: In addition to that, the comment in kfence.h needs to be updated to no longer mention 'size < 0' and 'size > 0'.

melver (changes requested):
Thanks!

mm/kfence/core.c
ramosian-glider: I broke this one manually, as they now say 80 chars is still the preferred limit. But maybe it doesn't make much sense to break a for-loop

ramosian-glider: ... especially given another for-loop crossing the 80-col limit a couple of lines above.

mm/kfence/core.c
ramosian-glider: Done

melver (approved):
LGTM, thanks!

PR #146, opened 2020-09-11
More fixes for v2, address long-lived objects in special caches

melver:
This seems to work really well, at least for the sysbench I/O workload. My guess is that most of these pathological allocations should all be via special caches.

### DIFF
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -219,15 +219,6 @@ static inline void for_each_canary(const struct kfence_metadata *meta, bool (*fn
 
 static void *kfence_guarded_alloc(struct kmem_cache *cache, size_t size, gfp_t gfp)
 {
-	/*
-	 * Note: for allocations made before RNG initialization, will always
-	 * return zero. We still benefit from enabling KFENCE as early as
-	 * possible, even when the RNG is not yet available, as this will allow
-	 * KFENCE to detect bugs due to earlier allocations. The only downside
-	 * is that the out-of-bounds accesses detected are deterministic for
-	 * such allocations.
-	 */
-	const bool right = prandom_u32_max(2);
 	struct kfence_metadata *meta = NULL;
 	unsigned long flags;
 	void *addr;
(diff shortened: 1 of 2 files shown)


dvyukov:
This looks premature to me. It maybe like speeding up a benchmark, but slowing down production workload. Custom caches have lots of transient allocations and kmalloc caches can well have persistent allocations...

mm/kfence/core.c
```diff
@@ -217,6 +223,28 @@ static inline void for_each_canary(const struct kfence_metadata *meta, bool (*fn
...
+		return false;
+
+	pool_usage = (100 * atomic_long_read(&counters[KFENCE_COUNTER_ALLOCATED])) /
+		     CONFIG_KFENCE_NUM_OBJECTS;
+	if (pool_usage > min(kfence_non_kmalloc_cutoff, 100U)) {
+		/*
+		 * Iterating through all kmem_caches is more expensive; just
+		 * checking the prefix of the name is simpler.
+		 */
+		if (!str_has_prefix(cache->name, "kmalloc"))
```
ramosian-glider, line 241: Note that there are "dma-kmalloc-XXX" caches as well.

melver:
> This looks premature to me. It maybe like speeding up a benchmark, but slowing down production workload. Custom caches have lots of transient allocations and kmalloc caches can well have persistent allocations...

This is not related to performance. It didn't speed up the benchmark. It ensured that it doesn't exhaust the memory pool.

The fact is still, that many of those long-lived allocation (like for I/O caches) come from custom caches. The cut-off also only kicks in after we reach e.g. 80% of usage, so we still get plenty of KFENCE allocations for those custom caches.

ramosian-glider (commented):
I have mixed feelings about these patches. It will be hard to remove this knob once people start depending on it. Also, despite kmalloc caches are more generic, I don't think we have enough data to say that most bugs are in the code using kmalloc.

mm/kfence/core.c
melver: If we've exhausted 80% of our pool, I'd say we're fine just servicing normal kmalloc(), and can safely ignore DMA things.

melver:
> I have mixed feelings about these patches. It will be hard to remove this knob once people start depending on it. Also, despite kmalloc caches are more generic, I don't think we have enough data to say that most bugs are in the code using kmalloc.

This makes sense, and is something I can get behind.

I just wanted to throw it out there, since we were discussing it and this is one option. And given the feedback, it seems we agree we can live without special tweaks and hacks for long-lived objects for now. :-)

PR #147, opened 2020-09-11
Minor fixes for v2

melver:
There were 2 smaller fixes in the other PR. Specifically the tweak to allow changing sample_interval for all builds at runtime, and calling prandom_u32() only if needed.

I am somewhat worried that changing sample_interval at runtime can be abused (e.g. by curious sysadmins), but I suppose if you're root, you already own the system anyway.

Thanks,
-- Marco


ramosian-glider (approved):
LGTM, thanks!

dvyukov (approved):

PR #149, opened 2020-09-16
kfence: v3: Lower line buffer size

melver:
To avoid warnings of using more than 1024 bytes stack usage.

In a test run with a WARN_ON(256 > len) no warnings were triggered for
the test, and 256 bytes should still be enough.


ramosian-glider (approved):
LGTM

PR #151, opened 2020-09-29
kfence: v4: Fix report generation if __slab_free tail-called

melver:
Found with Clang + linux-next + KFENCE.


ramosian-glider (approved):
LGTM, thanks!

PR #155, opened 2020-10-02
Some fixes for v5

melver:


arch/arm64/include/asm/kfence.h
```diff
@@ -17,17 +13,7 @@
...
-	const unsigned int num_pages = ilog2(roundup_pow_of_two(KFENCE_POOL_SIZE / PAGE_SIZE));
-	struct page *pages = alloc_pages(GFP_KERNEL, num_pages);
-
-	if (!pages)
-		return false;
-
-	__kfence_pool = page_address(pages);
-	return true;
-}
+static inline bool arch_kfence_initialize_pool(void) { return true; }
```
ramosian-glider, line 16: We probably need to declare the same stub for x86?

mm/kfence/core.c
```diff
@@ -414,6 +414,10 @@ static bool __init kfence_initialize_pool(void)
 		if (!i || (i % 2))
 			continue;
 
+		/* Verify we do not have a compound head page. */
```
ramosian-glider, line 418: According to your own reply to Jann, "In slub.c, __GFP_COMP is passed to alloc_pages(), which causes them to
have a compound head I believe."
How come we don't have a compound head here then?

arch/arm64/include/asm/kfence.h
melver: The function name didn't change. For x86, arch_kfence_initialize_pool() still has to do the set_memory_4k stuff on the static pool. 

Or maybe I missunderstood?

The way it should work:
- If we use the static pool, the generic alloc_kfence_pool() in core.c does nothing, as is the case for x86 right now.

- If we use the dynamic pool, one of 3 things can happen:
  1. arch_kfence_initialize_pool() is empty, and alloc_kfence_pool() just does the work.
  2. arch_kfence_initialize_pool() uses alloc_kfence_pool() (since it's included, it can declare the static function and just use it) and does some extra stuff.
  3. arch_kfence_initialize_pool() does something entirely different, and alloc_kfence_pool() will do nothing since __kfence_pool is non-NULL.

If we switch x86 over to a dynamic pool and give up on the static pool,I think it'll be option (ii) for x86.

mm/kfence/core.c
```diff
@@ -548,6 +548,14 @@ static void toggle_allocation_gate(struct work_struct *work)
 	if (!READ_ONCE(kfence_enabled))
 		return;
 
+	/*
```
ramosian-glider, line 551: I would have probably put this comment before the function, but don't have a strong preference.

mm/kfence/core.c
melver: Because we do not pass __GFP_COMP to alloc_pages() for our KFENCE pool.

ramosian-glider: I see, I've misunderstod that comment.

arch/arm64/include/asm/kfence.h
ramosian-glider: Ah, I've been staring at that function, but failed to understand it has the same name. Sorry about that!

mm/kfence/core.c
melver: Done.

melver:
PTAL.

include/linux/kfence.h
```diff
@@ -40,7 +40,7 @@ extern struct static_key_false kfence_allocation_key;
 static __always_inline bool is_kfence_address(const void *addr)
 {
 	return unlikely((char *)addr >= __kfence_pool &&
```
ramosian-glider: Doesn't this check guarantee that `addr` is non-null?

mm/kfence/core.c
```diff
@@ -233,16 +233,19 @@ static inline bool check_canary_byte(u8 *addr)
...
-	for (addr = ALIGN_DOWN(meta->addr, PAGE_SIZE); addr < meta->addr; addr++) {
+	/* Check left of object. */
+	for (addr = pageaddr; addr < meta->addr; addr++) {
 		if (!fn((u8 *)addr))
 			break;
 	}
 
-	for (addr = meta->addr + meta->size; addr < PAGE_ALIGN(meta->addr); addr++) {
+	/* Check right of object. */
+	for (addr = meta->addr + meta->size; addr < pageaddr + PAGE_SIZE; addr++) {
```
ramosian-glider, line 248: Nice catch, looks like for page-aligned meta->addr both ALIGN_DOWN and PAGE_ALIGN returned the same value.

include/linux/kfence.h
```diff
@@ -39,8 +39,7 @@ extern struct static_key_false kfence_allocation_key;
  */
 static __always_inline bool is_kfence_address(const void *addr)
 {
-	return unlikely((char *)addr >= __kfence_pool &&
-			(char *)addr < __kfence_pool + KFENCE_POOL_SIZE && addr);
+	return unlikely((unsigned long)((char *)addr - __kfence_pool) < KFENCE_POOL_SIZE && addr);
```
ramosian-glider, line 42: Ok, now checking for `addr` makes sense. But maybe it should look like: `likely(addr) && unlikely((unsigned long)((char *)addr - __kfence_pool) < KFENCE_POOL_SIZE)` then?

include/linux/kfence.h
melver: Jann's concern was that if __kfence_pool's pointer is NULL because KFENCE was never initialized.

include/linux/kfence.h
melver: The non-NULL check is still needed in case KFENCE was never initialized.

I think we want to write this so that the non-NULL check is in the unlikely slow-path:
```
Also, you could do the NULL *after* the range check said "true". That
way the NULL check would be on the slowpath and have basically no
performance impact.
```

melver: Also, hopefully by only having 1 branch for the range check, we train the CPU's branch predictor enough so that it'll predict the range check as false and we do not even notice the pointer load.

melver: Added a comment.

ramosian-glider: Ok, makes sense.

ramosian-glider (approved):
LGTM

PR #157, opened 2020-10-07
More fixes for v5, and removal of static pool

melver:
Some minor fixes, but major change here is removal of static pool.

After some new rounds of benchmarking, I find that there really isn't a noticeable difference between the 2 versions anymore. See commit log for more details.


ramosian-glider (approved):
Rest in peace, static pool.

Indeed, I also did some benchmarking of `perf bench sched messaging -l 30000` (3 rounds of 10 samples each, with reboots between rounds) and got the following results:

Static pool:
kfence.sample_interval=0: median 62.2915 sec, CV 1.60672483%
kfence.sample_interval=1000: median 62.3345 sec, CV 1.093313866%

Dynamic pool:
kfence.sample_interval=0: median 62.429 sec, CV 1.019534862%
kfence.sample_interval=1000: median 62.909 sec, CV 1.306502335%

The value of kfence.sample_interval=0 shouldn't really matter, as we're measuring the cost of replacing `sub    $0xffffffff82bd9000,%rax` (static pool, no load) with `sub    0x110c976(%rip),%rax` (dynamic pool, cached load) on the `kfree()` fast path.

PR #160, opened 2020-10-09
More fixes for v5

melver:
- Changes to KFENCE_NUM_OBJECTS handling. FYI: Turns out actual max order is MAX_ORDER-1 (🤔). So in reality the max number of objects on e.g. x86 is only 511.

- New KASAN problems, because of using page_alloc.

- set_direct_map_{invalid,default}_noflush() does not work due to potential deadlock.


ramosian-glider:
LGTM (sorry, couldn't log in from phone)

PR #161, opened 2020-10-14
Fix and memblock changes

melver:


melver:
arm64 doesn't work yet.

melver:
With this latest version, both x86 and arm64 seem to work. Still running more tests.

melver:
Survived syzkaller testing as well with varying pool sizes.

ramosian-glider (approved):
LGTM

PR #163, opened 2020-10-21
kfence: More useful and graceful shutdown_cache() behavior

melver:
See first commit message for description. Also fixes SLAB_TYPESAFE_BY_RCU handling, which was incorrect.

### DIFF
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -59,13 +59,22 @@ void __init kfence_init(void);
  * kfence_shutdown_cache() - handle shutdown_cache() for KFENCE objects
  * @s: cache being shut down
  *
- * Return: true on success, false if any leftover objects persist.
- *
  * Before shutting down a cache, one must ensure there are no remaining objects
- * allocated from it. KFENCE objects are not referenced from the cache, so
- * kfence_shutdown_cache() takes care of them.
+ * allocated from it. Because KFENCE objects are not referenced from the cache
+ * directly, we need to check them here.
+ *
+ * Note that shutdown_cache() is internal to SL*B, and kmem_cache_destroy() does
+ * not return if allocated objects still exist: it prints an error message and
+ * simply aborts destruction of a cache, leaking memory.
+ *
+ * If the only such objects are KFENCE objects, we will not leak the entire
+ * cache, but instead try to provide more useful debug info by making allocated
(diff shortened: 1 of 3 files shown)


include/linux/kfence.h
```diff
@@ -59,13 +59,22 @@ void __init kfence_init(void);
...
- *
  * Before shutting down a cache, one must ensure there are no remaining objects
- * allocated from it. KFENCE objects are not referenced from the cache, so
- * kfence_shutdown_cache() takes care of them.
+ * allocated from it. Because KFENCE objects are not referenced from the cache
+ * directly, we need to check them here.
+ *
+ * Note that shutdown_cache() is internal to SL*B, and kmem_cache_destroy() does
+ * not return if allocated objects still exist: it prints an error message and
+ * simply aborts destruction of a cache, leaking memory.
```
ramosian-glider, line 68: I thought we could maybe just call kfence_shutdown_cache() after __kmem_cache_shutdown(), so that we mimic the default kernel behavior, but this approach is better.

mm/slab_common.c
```diff
@@ -449,9 +450,6 @@ static int shutdown_cache(struct kmem_cache *s)
 	/* free asan quarantined objects */
 	kasan_cache_shutdown(s);
 
-	if (!kfence_shutdown_cache(s))
```
ramosian-glider, line 452: Ah, ok, that is effectively what you are doing - moving this call after __kmem_cache_shutdown, so that we'll leak the whole cache in the case of an early exit from shutdown_cache(), correct?

ramosian-glider (approved):
LGTM

mm/slab_common.c
melver: Yes, if there are non-KFENCE objects there's nothing we can do. But if there are only KFENCE objects, we'll let the kernel release the cache normally, we just leak some KFENCE objects by zombie'fying them. We could maybe do better to provide helpful error messages by calling kfence_shutdown_cache() in all cases (including early return due to __kmem_cache_shutdown()), but with the whole SLAB_TYPESAFE_BY_RCU business this becomes really ugly and is way more than the 2 calls to kfence_shutdown_cache() we have now.

The current version is safe for RCU and non-RCU cases, but maybe we won't get very helpful error reports if there are other non-KFENCE objects still allocated.

Issue #164, opened 2020-10-21
kfence: Debug memcg issues


melver:
Currently it seems that KFENCE has problems with memcg allocations. To reproduce, 

```
CONFIG_MEMCG=y
CONFIG_MEMCG_SWAP=y
CONFIG_MEMCG_KMEM=y
```

Then use `tools/testing/selftests/cgroup/test_kmem` to test. (Hint: boot with cgroup_no_v1=all on old distros, then `mount -t cgroup2 none /sys/fs/cgroup`).

For example (with SLAB):
```
general protection fault, probably for non-canonical address 0x6aba3f0293b07967: 0000 [#1] PREEMPT SMP PTI
CPU: 2 PID: 469 Comm: test_kmem Not tainted 5.9.0-rc4+ #10
Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.13.0-1 04/01/2014
RIP: 0010:obj_cgroup_memcg include/linux/memcontrol.h:526 [inline]
RIP: 0010:mem_cgroup_from_obj+0x68/0x80 mm/memcontrol.c:2902
Code: 49 83 e0 fe 89 d8 8b 4a 18 48 0f af c1 0f b6 4a 1c 48 c1 e8 20 29 c3 d3 eb 0f b6 4a 1d 01 c3 d3 eb 4d 8b 04 d8 4d 85 c0 74 04 <4d> 8b 40 38 4c 89 c0 5b c3 45 31 c0 4c 89 c0 c3 0f 1f 84 00 00 00
RSP: 0018:ffffaada00267b80 EFLAGS: 00010202
RAX: 0000000008492ad5 RBX: 00000000002124ab RCX: 0000000000000007
RDX: ffff96b917965cc0 RSI: ffff96b918db8080 RDI: 0000000080000000
RBP: ffff96b914996c80 R08: 6aba3f0293b07967 R09: ffffaada00267ae0
R10: 000000000002cbb8 R11: 0000000000000005 R12: ffff96b915345a40
R13: 0000000000000000 R14: ffff96b915345a48 R15: ffff96b9166249c0
FS:  00007fd34d3ca700(0000) GS:ffff96b91fc80000(0000) knlGS:0000000000000000
CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
CR2: 00007fd34cccf280 CR3: 0000000816674003 CR4: 0000000000770ee0
[...]

PR #165, opened 2020-10-22
Fix kfence+memcg

melver:
Fixes https://github.com/google/kasan/issues/164

One thing I don't like is that we are adding an include to the *def.h headers. These seem to want no headers? Although technically they should include their required headers... hopefully nobody will complain about us adding the kfence header (the only major include in kfence.h is mm.h).

### DIFF
--- a/include/linux/slab_def.h
+++ b/include/linux/slab_def.h
@@ -2,6 +2,7 @@
 #ifndef _LINUX_SLAB_DEF_H
 #define	_LINUX_SLAB_DEF_H
 
+#include <linux/kfence.h>
 #include <linux/reciprocal_div.h>
 
 /*
@@ -114,6 +115,8 @@ static inline unsigned int obj_to_index(const struct kmem_cache *cache,
 static inline int objs_per_slab_page(const struct kmem_cache *cache,
 				     const struct page *page)
 {
+	if (is_kfence_address(page_address(page)))
+		return 1;
 	return cache->num;
 }
 
--- a/include/linux/slub_def.h
+++ b/include/linux/slub_def.h
@@ -7,6 +7,7 @@
  *
  * (C) 2007 SGI, Christoph Lameter
  */
+#include <linux/kfence.h>
 #include <linux/kobject.h>
 #include <linux/reciprocal_div.h>
 
@@ -185,6 +186,8 @@ static inline unsigned int __obj_to_index(const struct kmem_cache *cache,
 static inline unsigned int obj_to_index(const struct kmem_cache *cache,
 					const struct page *page, void *obj)
 {
+	if (is_kfence_address(obj))
+		return 0;
 	return __obj_to_index(cache, page_address(page), obj);
 }
 
--- a/include/linux/kfence.h
+++ b/include/linux/kfence.h
@@ -4,7 +4,6 @@
 #define _LINUX_KFENCE_H
 
 #include <linux/mm.h>
-#include <linux/percpu.h>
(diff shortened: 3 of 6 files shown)


mm/slab.h
```diff
@@ -324,7 +325,7 @@ static inline void memcg_slab_post_alloc_hook(struct kmem_cache *s,
 
 	flags &= ~__GFP_ACCOUNT;
 	for (i = 0; i < size; i++) {
-		if (likely(p[i])) {
+		if (likely(p[i] && !is_kfence_address(p[i]))) {
```
ramosian-glider: We probably need a comment here about why KFENCE objects do not need to be charged.

mm/slab.h
```diff
@@ -325,7 +324,7 @@ static inline void memcg_slab_post_alloc_hook(struct kmem_cache *s,
 
 	flags &= ~__GFP_ACCOUNT;
 	for (i = 0; i < size; i++) {
-		if (likely(p[i] && !is_kfence_address(p[i]))) {
+		if (likely(p[i])) {
```
ramosian-glider: I should learn to read the whole series before writing comments.

mm/slab.h
melver: This commit is being reverted. :-) 

I just left it here for keeping the history.

include/linux/slab_def.h
```diff
@@ -114,6 +115,8 @@ static inline unsigned int obj_to_index(const struct kmem_cache *cache,
 static inline int objs_per_slab_page(const struct kmem_cache *cache,
 				     const struct page *page)
 {
+	if (is_kfence_address(page_address(page)))
```
ramosian-glider, line 118: What if we just move this check to memcg_alloc_page_obj_cgroups(), which seems to be the only caller of objs_per_slab_page? This way we'll avoid adding a dependency on kfence.h

include/linux/slub_def.h
```diff
@@ -185,6 +186,8 @@ static inline unsigned int __obj_to_index(const struct kmem_cache *cache,
 static inline unsigned int obj_to_index(const struct kmem_cache *cache,
 					const struct page *page, void *obj)
 {
+	if (is_kfence_address(obj))
```
ramosian-glider, line 189: Hm, the same could be done here, but there are multiple callers of obj_to_index(), and we probably don't want to open-code them all.
I wonder if this doesn't violate any contracts about the index being translatable back to the object address.

include/linux/slab_def.h
melver: If we do that we can't easily test this function in kfence_test. There may also be future callers and things will break.

I'm willing to accept the include for sake of saner code. If we get push-back, we can either change it, or argue that this file should include its dependencies anyway (I actually don't know how the thing still works, given it refers to things that it doesn't include nor mm/slab.h includes -- probably by accident?! It's a mess, let's not make it messier ;-)).

include/linux/slub_def.h
melver: We should push for clean code first. If somebody wants ugly code, we can make it ugly. ;-)

Re violating contracts: For slab.c index_to_obj() works because we set s_mem correctly. Other than those in slab.c, I don't see any reverse translations.

And I also see we can probably get rid of the is_kfence_address() guard in slab.c's __check_heap_object() for the obj_to_index() translation (it just returns 0). I'll add that commit. So evidently, things before these fixes were definitely more broken.

ramosian-glider (approved):
LGTM

PR #166, opened 2020-10-27
kfence: Record allocation and free task pids

melver:
This can either go into a v6, or as a separate patch later.

### DIFF
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -182,19 +182,17 @@ static inline unsigned long metadata_to_pageaddr(const struct kfence_metadata *m
 static noinline void metadata_update_state(struct kfence_metadata *meta,
 					   enum kfence_object_state next)
 {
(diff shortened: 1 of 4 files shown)


mm/kfence/core.c
```diff
@@ -191,10 +191,13 @@ static noinline void metadata_update_state(struct kfence_metadata *meta,
 
 	lockdep_assert_held(&meta->lock);
 
-	if (next == KFENCE_OBJECT_FREED)
+	if (next == KFENCE_OBJECT_FREED) {
 		meta->num_free_stack = nentries;
-	else
+		meta->free_task_pid = task_pid_nr(current);
```
ramosian-glider, line 196: Maybe it's time to factor alloc/free metadata into a separate struct?

ramosian-glider (approved):

mm/kfence/core.c
melver: Ah yes of course, done!

ramosian-glider (commented):
LGTM++

PR #169, opened 2020-11-04
kfence: Use pt_regs to generate stack trace on faults

melver:
To be sent as individual patch to LKML.

### DIFF
--- a/mm/kfence/report.c
+++ b/mm/kfence/report.c
@@ -5,6 +5,7 @@
 #include <linux/kernel.h>
 #include <linux/lockdep.h>
 #include <linux/printk.h>
+#include <linux/sched/debug.h>
 #include <linux/seq_file.h>
 #include <linux/stacktrace.h>
 #include <linux/string.h>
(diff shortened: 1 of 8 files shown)


mm/kfence/report.c
```diff
@@ -222,7 +228,10 @@ void kfence_report_error(unsigned long address, const struct kfence_metadata *me
 
 	/* Print report footer. */
 	pr_err("\n");
-	dump_stack_print_info(KERN_ERR);
+	if (IS_ENABLED(CONFIG_DEBUG_KERNEL) && regs)
+		show_regs(regs);
```
ramosian-glider, line 232: Does this function dump the stack to KERN_ERR or KERN_DEFAULT?

ramosian-glider (approved):
LGTM, thank you!

mm/kfence/report.c
melver: KERN_DEFAULT on x86 and arm64. I think our report is currently still split between err and default (I think our stack traces are DEFAULT, whereas title+separators is ERR, etc.). As long as dmesg contains the full report it's fine. I think it matters only when printed lines are filtered on serial console or so (which I've already observed). But there we see at least the title.

PR #171, opened 2020-12-17
error_report_notify implementation

ramosian-glider:
This patch series adds error report tracing for debugging tools like KFENCE and KASAN, so that it is possible to extend their error reporting mechanisms.
It also implements CONFIG_ERROR_REPORT_NOTIFY that captures the last error report and exposes it, together with the report count, to the userspace via sysfs.
This shall make it possible to perform userspace handling of memory errors (e.g. uploading them to the cloud) in restrictive environments that disallow access to dmesg or /dev/kmsg.

### DIFF
--- a/include/trace/events/error_report.h
+++ b/include/trace/events/error_report.h
@@ -1,0 +1,51 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#undef TRACE_SYSTEM
+#define TRACE_SYSTEM error_report
+
+#if !defined(_TRACE_ERROR_REPORT_H) || defined(TRACE_HEADER_MULTI_READ)
+#define _TRACE_ERROR_REPORT_H
+
+#include <linux/tracepoint.h>
+
+DECLARE_EVENT_CLASS(error_report_template,
+		    TP_PROTO(const char *error_detector, unsigned long id),
+		    TP_ARGS(error_detector, id),
+		    TP_STRUCT__entry(__field(const char *, error_detector)
+					     __field(unsigned long, id)),
+		    TP_fast_assign(__entry->error_detector = error_detector;
+				   __entry->id = id;),
+		    TP_printk("[%s] %lx", __entry->error_detector,
+			      __entry->id));
+
+/**
+ * error_report_start - called before printing the error report
+ * @error_detector:	short string describing the error detection tool
+ * @id:			pseudo-unique descriptor that can help distinguish reports
+ * 			from one another. Depending on the tool, good examples
+ * 			could be: memory access address, call site, allocation
+ * 			site, etc.
+ *
+ * This event occurs right before a debugging tool starts printing the error
+ * report.
+ */
+DEFINE_EVENT(error_report_template, error_report_start,
+	     TP_PROTO(const char *error_detector, unsigned long id),
+	     TP_ARGS(error_detector, id));
+
+/**
+ * error_report_end - called after printing the error report
+ * @error_detector:	short string describing the error detection tool
+ * @id:			pseudo-unique descriptor, matches that passed to
+ * 			error_report_start
+ *
+ * This event occurs right after a debugging tool finishes printing the error
+ * report.
+ */
+DEFINE_EVENT(error_report_template, error_report_end,
+	     TP_PROTO(const char *error_detector, unsigned long id),
+	     TP_ARGS(error_detector, id));
+
+#endif /* _TRACE_ERROR_REPORT_H */
+
+/* This part must be outside protection */
+#include <trace/define_trace.h>
--- a/kernel/trace/error_report-traces.c
+++ b/kernel/trace/error_report-traces.c
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Error reporting trace points
+ */
+
+#define CREATE_TRACE_POINTS
+#include <trace/events/error_report.h>
+
+EXPORT_TRACEPOINT_SYMBOL_GPL(error_report_start);
+EXPORT_TRACEPOINT_SYMBOL_GPL(error_report_end);
+
--- a/lib/Kconfig.debug
+++ b/lib/Kconfig.debug
@@ -195,6 +195,20 @@ config DEBUG_BUGVERBOSE
 	  of the BUG call as well as the EIP and oops trace.  This aids
 	  debugging but costs about 70-100K of memory.
 
+config ERROR_REPORT_NOTIFY
+	bool "Expose memory error reports to the userspace"
+	depends on TRACING
+	help
+	  When enabled, captures error reports from debugging tools (such as
+	  KFENCE or KASAN) using console tracing, and exposes reports in
+	  /sys/kernel/error_report/: the file last_report contains the last
+	  report (with maximum report length of PAGE_SIZE), and report_count,
+	  the total report count.
+
+	  Userspace programs can call poll(POLLPRI) on those files to get
+	  notified about the new reports without having to watch dmesg in a
+	  loop.
+
 endmenu # "printk and dmesg options"
 
 menu "Compile-time checks and compiler options"
--- a/lib/error_report_notify.c
+++ b/lib/error_report_notify.c
@@ -1,0 +1,278 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/*
+ * Userspace notification interface for debugging tools.
+ *
+ * Provide two sysfs files:
+ *  - /sys/kernel/error_report/last_report
+ *  - /sys/kernel/error_report/report_count
+ * that contain the last debugging tool report (taken from dmesg, delimited by
+ * the error_report_start/error_report_end tracing events) and the total report
+ * count.
+ */
+
+#include <linux/atomic.h>
+#include <linux/fs.h>
+#include <linux/kobject.h>
+#include <linux/string.h>
+#include <linux/sysfs.h>
+#include <linux/tracepoint.h>
+#include <linux/workqueue.h>
+#include <trace/events/error_report.h>
+#include <trace/events/printk.h>
+
+static struct kobject *error_report_kobj;
+
+/* sysfs files are capped at PAGE_SIZE. */
+#define BUF_SIZE PAGE_SIZE
+/* Two buffers to store the finished report and the report being recorded. */
+static char report_buffer[2][BUF_SIZE];
+/*
+ * Total report count. Also serves as a latch for report_buffer:
+ * report_buffer[num_reports % 2] is the currently available report,
+ * report_buffer[(num_reports + 1) % 2] is the report being recorded.
+ */
+static atomic_t num_reports;
+
+/*
+ * PID of the task currently recording the report, as returned by
+ * get_encoded_pid(), or -1. Used as a writer lock for report_buffer.
+ * A regular spinlock couldn't be used here, as probe_console() can be called
+ * from any thread, and it needs to know whether that thread is holding the
+ * lock.
+ */
+static atomic_t current_pid = ATOMIC_INIT(-1);
+
+static size_t current_pos;
+static bool truncated;
+static const char TRUNC_MSG[] = "<truncated>\n";
+
+static struct delayed_work reporting_done;
+
+static void error_report_notify(struct work_struct *work)
+{
+	sysfs_notify(error_report_kobj, NULL, "last_report");
+	sysfs_notify(error_report_kobj, NULL, "report_count");
+}
+static DECLARE_DELAYED_WORK(reporting_done, error_report_notify);
+
+/*
+ * Return the current PID combined together with in_task(). This lets us
+ * distinguish between normal task context and IRQ context.
+ */
+static int get_encoded_pid(void)
+{
+	return (current->pid << 1) | (!!in_task());
+}
+
+/*
+ * Trace hook for the error_report_start event. In an unlikely case of another
+ * task already printing a report bail out, otherwise save the current pid
+ * together with in_task() return value.
+ *
+ * Because reporting code can be called from low-level routines (e.g. locking
+ * primitives or allocator guts), report recording is implemented using a
+ * seqlock lock-free algorithm.
+ */
+static void probe_report_start(void *ignore, const char *tool_name,
+			       unsigned long id)
+{
+	/*
+	 * Acquire the writer lock. Any racing probe_report_start will not
+	 * record anything. Pairs with the release in probe_report_end().
+	 */
+	if (atomic_cmpxchg_acquire(&current_pid, -1, get_encoded_pid()) != -1)
+		return;
+	current_pos = 0;
+	truncated = false;
+}
+
+/*
+ * Trace hook for the error_report_end event. If an event from the mismatching
+ * error_report_start is received, it is ignored. Otherwise, null-terminate the
+ * buffer, increase the report count (effectively releasing the report to
+ * last_report_show() and schedule a notification about a new report.
+ */
+static void probe_report_end(void *ignore, const char *tool_name,
+			     unsigned long id)
+{
+	pid_t pid = atomic_read(&current_pid);
+	int idx;
+
+	if (pid != get_encoded_pid())
+		return;
+
+	idx = (atomic_read(&num_reports) + 1) % 2;
+	if (current_pos == BUF_SIZE)
+		report_buffer[idx][current_pos - 1] = 0;
+	else
+		report_buffer[idx][current_pos] = 0;
+
+	/* Pairs with acquire in last_report_show(). */
+	atomic_inc_return_release(&num_reports);
+	schedule_delayed_work(&reporting_done, 0);
+	/*
+	 * Release the writer lock. Pairs with the acquire in
+	 * probe_report_start().
+	 */
+	atomic_set_release(&current_pid, -1);
+}
+
+/*
+ * Skip one or two leading pair of brackets containing the log timestamp and
+ * the task/CPU ID, plus the leading space, from the report line, e.g.:
+ *   [    0.698431][    T7] BUG: KFENCE: use-after-free ...
+ * becomes:
+ *   BUG: KFENCE: use-after-free ...
+ *
+ * Report size is only 4K, and this boilerplate can easily account for half of
+ * that amount.
+ */
+static void skip_extra_info(const char **buf, size_t *len)
+{
+	int num_brackets = IS_ENABLED(CONFIG_PRINTK_TIME) +
+			   IS_ENABLED(CONFIG_PRINTK_CALLER);
+	const char *found;
+
+	if (!buf || !len)
+		return;
+
+	while (num_brackets--) {
+		if (!*len || *buf[0] != '[')
+			return;
+		found = strnchr(*buf, *len, ']');
+		if (!found)
+			return;
+		*len -= found - *buf + 1;
+		*buf = found + 1;
+	}
+	if (*len && *buf[0] == ' ') {
+		++*buf;
+		--*len;
+	}
+}
+
+/*
+ * Trace hook for the console event. If a line comes from a task/CPU that did
+ * not send the error_report_start event, that line is ignored. Otherwise, it
+ * is stored in the report_buffer[(num_reports + 1) % 2].
+ *
+ * To save space, the leading timestamps and (when enabled) CPU/task info is
+ * stripped away. The buffer may contain newlines, so this procedure is
+ * repeated for every line.
+ */
+static void probe_console(void *ignore, const char *buf, size_t len)
+{
+	int pid = atomic_read(&current_pid);
+	size_t to_copy, cur_len;
+	char *newline;
+	int idx;
+
+	if (pid != get_encoded_pid() || truncated)
+		return;
+
+	idx = (atomic_read(&num_reports) + 1) % 2;
+	while (len) {
+		newline = strnchr(buf, len, '\n');
+		if (newline)
+			cur_len = newline - buf + 1;
+		else
+			cur_len = len;
+		/* Adjust len now, because skip_extra_info() may change cur_len. */
+		len -= cur_len;
+		skip_extra_info(&buf, &cur_len);
+		to_copy = min(cur_len, BUF_SIZE - current_pos);
+		memcpy(report_buffer[idx] + current_pos, buf, to_copy);
+		current_pos += to_copy;
+		if (cur_len > to_copy) {
+			truncated = true;
+			memcpy(report_buffer[idx] + current_pos - sizeof(TRUNC_MSG),
+			       TRUNC_MSG, sizeof(TRUNC_MSG));
+			break;
+		}
+		buf += cur_len;
+	}
+}
+
+static void register_tracepoints(void)
+{
+	register_trace_console(probe_console, NULL);
+	register_trace_error_report_start(probe_report_start, NULL);
+	register_trace_error_report_end(probe_report_end, NULL);
+}
+
+/*
+ * read() handler for /sys/kernel/error_report/last_report.
+ * Because the number of reports can change under our feet, check it again
+ * after copying the report, and retry if the numbers mismatch.
+ * */
+static ssize_t last_report_show(struct kobject *kobj,
+				struct kobj_attribute *attr, char *buf)
+{
+	ssize_t ret;
+	int index;
+
+	do {
+		/* Pairs with release in probe_report_end(). */
+		index = atomic_read_acquire(&num_reports);
+		/*
+		 * If index and old_index mismatch, we might be accessing
+		 * report_buffer concurrently with a writer thread. In that
+		 * case the read data will be discarded.
+		 */
+		ret = data_race(strscpy(buf, report_buffer[index % 2], BUF_SIZE));
+		/*
+		 * Prevent reordering between the memcpy above and the atomic
+		 * read below.
+		 * See the comments in include/linux/seqlock.h for more
+		 * details.
+		 */
+		smp_rmb();
+	} while (index != atomic_read(&num_reports));
+	return ret;
+}
+
+/*
+ * read() handler for /sys/kernel/error_report/report_count.
+ */
+static ssize_t report_count_show(struct kobject *kobj,
+				 struct kobj_attribute *attr, char *buf)
+{
+	return scnprintf(buf, PAGE_SIZE, "%d\n", atomic_read(&num_reports));
+}
+
+static struct kobj_attribute last_report_attr = __ATTR_RO(last_report);
+static struct kobj_attribute report_count_attr = __ATTR_RO(report_count);
+static struct attribute *error_report_sysfs_attrs[] = {
+	&last_report_attr.attr,
+	&report_count_attr.attr,
+	NULL,
+};
+
+static const struct attribute_group error_report_sysfs_attr_group = {
+	.attrs = error_report_sysfs_attrs,
+};
+
+/*
+ * Set up report notification: register tracepoints and create
+ * /sys/kernel/error_report/.
+ */
+static void error_report_notify_setup(void)
+{
+	int err;
+
+	register_tracepoints();
+	error_report_kobj = kobject_create_and_add("error_report", kernel_kobj);
+	if (!error_report_kobj)
+		goto error;
+	err = sysfs_create_group(error_report_kobj,
+				 &error_report_sysfs_attr_group);
+	if (err)
+		goto error;
+	return;
+
+error:
+	if (error_report_kobj)
+		kobject_del(error_report_kobj);
+}
+late_initcall(error_report_notify_setup);
--- a/mm/kasan/report.c
+++ b/mm/kasan/report.c
@@ -30,6 +30,7 @@
 #include <linux/module.h>
 #include <linux/sched/task_stack.h>
 #include <linux/uaccess.h>
+#include <trace/events/error_report.h>
 
 #include <asm/sections.h>
 
@@ -80,7 +81,7 @@ static void print_error_description(struct kasan_access_info *info)
 
 static DEFINE_SPINLOCK(report_lock);
 
-static void start_report(unsigned long *flags)
+static void start_report(unsigned long *flags, unsigned long addr)
 {
 	/*
 	 * Make sure we don't end up in loop.
@@ -88,10 +89,12 @@ static void start_report(unsigned long *flags)
 	kasan_disable_current();
 	spin_lock_irqsave(&report_lock, *flags);
 	pr_err("==================================================================\n");
+	trace_error_report_start("kasan", addr);
 }
 
-static void end_report(unsigned long *flags)
+static void end_report(unsigned long *flags, unsigned long addr)
 {
+	trace_error_report_end("kasan", addr);
 	pr_err("==================================================================\n");
 	add_taint(TAINT_BAD_PAGE, LOCKDEP_NOW_UNRELIABLE);
 	spin_unlock_irqrestore(&report_lock, *flags);
@@ -497,14 +500,14 @@ void kasan_report_invalid_free(void *object, unsigned long ip)
 		kasan_update_kunit_status(current->kunit_test);
 #endif /* IS_ENABLED(CONFIG_KUNIT) */
 
-	start_report(&flags);
+	start_report(&flags, (unsigned long)object);
 	pr_err("BUG: KASAN: double-free or invalid-free in %pS\n", (void *)ip);
 	print_tags(tag, object);
 	pr_err("\n");
 	print_address_description(object, tag);
 	pr_err("\n");
 	print_shadow_for_address(object);
-	end_report(&flags);
+	end_report(&flags, (unsigned long)object);
 }
 
 static void __kasan_report(unsigned long addr, size_t size, bool is_write,
@@ -534,7 +537,7 @@ static void __kasan_report(unsigned long addr, size_t size, bool is_write,
 	info.is_write = is_write;
 	info.ip = ip;
 
-	start_report(&flags);
+	start_report(&flags, addr);
 
 	print_error_description(&info);
 	if (addr_has_shadow(untagged_addr))
@@ -549,7 +552,7 @@ static void __kasan_report(unsigned long addr, size_t size, bool is_write,
 		dump_stack();
 	}
 
-	end_report(&flags);
+	end_report(&flags, addr);
 }
 
 bool kasan_report(unsigned long addr, size_t size, bool is_write,
--- a/mm/kfence/report.c
+++ b/mm/kfence/report.c
@@ -8,6 +8,7 @@
 #include <linux/seq_file.h>
 #include <linux/stacktrace.h>
 #include <linux/string.h>
+#include <trace/events/error_report.h>
 
 #include <asm/kfence.h>
 
@@ -178,6 +179,7 @@ void kfence_report_error(unsigned long address, const struct kfence_metadata *me
 	lockdep_off();
 
 	pr_err("==================================================================\n");
+	trace_error_report_start("kfence", address);
 	/* Print report header. */
 	switch (type) {
 	case KFENCE_ERROR_OOB: {
@@ -223,6 +225,7 @@ void kfence_report_error(unsigned long address, const struct kfence_metadata *me
 	/* Print report footer. */
 	pr_err("\n");
 	dump_stack_print_info(KERN_ERR);
+	trace_error_report_end("kfence", address);
 	pr_err("==================================================================\n");
 
 	lockdep_on();
--- a/kernel/trace/Makefile
+++ b/kernel/trace/Makefile
@@ -81,6 +81,7 @@ obj-$(CONFIG_SYNTH_EVENTS) += trace_events_synth.o
 obj-$(CONFIG_HIST_TRIGGERS) += trace_events_hist.o
 obj-$(CONFIG_BPF_EVENTS) += bpf_trace.o
 obj-$(CONFIG_KPROBE_EVENTS) += trace_kprobe.o
+obj-$(CONFIG_TRACEPOINTS) += error_report-traces.o
 obj-$(CONFIG_TRACEPOINTS) += power-traces.o
 ifeq ($(CONFIG_PM),y)
 obj-$(CONFIG_TRACEPOINTS) += rpm-traces.o
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -239,6 +239,8 @@ obj-$(CONFIG_CPU_RMAP) += cpu_rmap.o
 
 obj-$(CONFIG_DQL) += dynamic_queue_limits.o
 
+obj-$(CONFIG_ERROR_REPORT_NOTIFY) += error_report_notify.o
+
 obj-$(CONFIG_GLOB) += glob.o
 obj-$(CONFIG_GLOB_SELFTEST) += globtest.o
 


kernel/trace/error_report-traces.c
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Error reporting trace points
+ *
```
ramosian-glider, line 4: Spare line, will remove.

include/trace/events/error_report.h
```diff
@@ -0,0 +1,30 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
```
melver, line 1: Commit-message: it's not just restricted to "memory tools", it's more generic. I also wouldn't mention KCSAN here, because it raises the expectation that we'll add support in this series, which I think we shouldn't for now (I'll do that at some point, but not sure yet).

I think we might actually get some value out of adding the tracepoints to UBSAN as well, because UBSAN can be enabled in prod. So maybe the list of examples should be "KASAN, KFENCE, and UBSAN".

The description should also motivate better why this is useful standalone, without even considering error_report_notify. I would mention that it allows tests to hook into arbitrary bug and error reporting. And it also allows to benefit from ftrace event tracing.

include/trace/events/error_report.h
```diff
@@ -0,0 +1,30 @@
...
+#undef TRACE_SYSTEM
+#define TRACE_SYSTEM error_report
+
+#if !defined(_TRACE_ERROR_REPORT_H) || defined(TRACE_HEADER_MULTI_READ)
+#define _TRACE_ERROR_REPORT_H
+
+#include <linux/tracepoint.h>
+
+DECLARE_EVENT_CLASS(error_report_template,
```
melver, line 10: These need doc-comments. E.g. see include/trace/events/workqueue.h

include/trace/events/error_report.h
```diff
@@ -0,0 +1,30 @@
...
+		    TP_PROTO(const char *tool_name, unsigned long id),
```
melver, line 11: Instead of 'tool_name', would just 'error_detector' be more descriptive? (Without _name, as it's almost implicit from const char* type.)

mm/kfence/report.c
```diff
@@ -177,6 +178,7 @@ void kfence_report_error(unsigned long address, const struct kfence_metadata *me
 	 */
 	lockdep_off();
 
+	trace_error_report_start("kfence", address);
```
melver, line 181: Do we want the report with or without the '==============='.

I think it should be without, because it doesn't add much value in the file and just eats space from the buffer.

mm/kasan/report.c
```diff
@@ -80,19 +81,21 @@ static void print_error_description(struct kasan_access_info *info)
...
 
-static void start_report(unsigned long *flags)
+static void start_report(unsigned long *flags, unsigned long addr)
 {
 	/*
 	 * Make sure we don't end up in loop.
 	 */
 	kasan_disable_current();
 	spin_lock_irqsave(&report_lock, *flags);
+	trace_error_report_start("kasan", addr);
```
melver, line 91: Same here, with or without '==================='?

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
+// SPDX-License-Identifier: GPL-2.0
```
melver, line 1: Commit-message: I think it needs to be better motivated (and don't mention KCSAN, since that's not happening in this series).

I would also strengthen the motivation by copying bits from the design doc:
```
With the introduction of various production error-detection tools, such as
MTE-based KASAN and KFENCE, the need arises to efficiently notify user space
tools about kernel errors. Currently, no facility exists to notify user space
about a kernel error from such bug-detection tools. The problem is obviously
not restricted to the above bug detection tools, and applies to any error
reporting mechanism that does not panic the kernel; this series, however, will
only add support for KASAN and KFENCE reporting.

All such error reports appear in the kernel log. But, when such errors
occur, user space would normally need to read the entire kernel log and parse
the relevant errors. This is error prone and inefficient, as user space needs
to continuously monitor the kernel log and parse errors. On certain devices,
this is unfortunately not acceptable. Therefore, we need to revisit how reports
are propagated to user space.

The library added, error_report_notify (CONFIG_ERROR_REPORT_NOTIFY),
solves the above by using the error_report_start/error_report_end ...<rest of text>...
```

Minor:
- the config does not collect errors. The library does.

lib/Kconfig.debug
```diff
@@ -195,6 +195,21 @@ config DEBUG_BUGVERBOSE
...
 	  debugging but costs about 70-100K of memory.
 
+config ERROR_REPORT_NOTIFY
+	bool "Expose memory error reports to the userspace"
+	depends on TRACING
+	default n
+	help
+	  When enabled, captures error reports from debugging tools like KFENCE
+	  or KASAN using console tracing, and provides two files:
+	   - /sys/kernel/error_report/last_report
```
melver, line 205: I think the formatting when using:
```
- ...
- ...
```
is bad when using it with Kconfig help text. Maybe just:
```
When enabled, captures error reports from debugging tools (such as KFENCE
and KASAN) via console tracing, and exposes reports in /sys/kernel/error_report:
the file last_report contains the last report, and report_count the total number of
reports.
<rest about user space programs>
```
I think mentioning PAGE_SIZE is irrelevant, because it's not anything that a user can act on, and most likely will forget anyway.

A better UX would be to copy '<trimmed>' at the end of reports that are actually trimmed.

lib/Kconfig.debug
```diff
@@ -195,6 +195,21 @@ config DEBUG_BUGVERBOSE
 	  of the BUG call as well as the EIP and oops trace.  This aids
 	  debugging but costs about 70-100K of memory.
 
+config ERROR_REPORT_NOTIFY
+	bool "Expose memory error reports to the userspace"
+	depends on TRACING
+	default n
```
melver, line 201: Remove 'default n'

melver (changes requested):

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+#include <linux/string.h>
+#include <linux/tracepoint.h>
+#include <linux/workqueue.h>
+#include <trace/events/error_report.h>
+#include <trace/events/printk.h>
+
+static struct kobject *error_report_kobj;
+static atomic_t num_error_reports;
+
+static char current_report[PAGE_SIZE], last_report[PAGE_SIZE];
```
melver, line 28: Why not just make this a `buffered_report[2][PAGE_SIZE]' and have an int 'current_report_idx' and then on report completion just current_report_idx ^= 1 that?

It'll save all the memcpys.

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+{
+	unsigned long flags;
+	pid_t pid = atomic_read_acquire(&current_pid);
+
+	if (!current || pid != current->pid)
+		return;
+	spin_lock_irqsave(&last_report_lock, flags);
+	if (current_report_pos)
+		memcpy(last_report, current_report, current_report_pos);
+	last_report_size = current_report_pos;
```
melver, line 69: last_report_size isn't required elsewhere, remove?

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+static ssize_t last_report_show(struct kobject *kobj,
+				struct kobj_attribute *attr, char *buf)
+{
+	ssize_t ret;
+	unsigned long flags;
+
+	spin_lock_irqsave(&last_report_lock, flags);
+	ret = scnprintf(buf, PAGE_SIZE, last_report);
```
melver, line 111: Why does it have to use scnprintf? Can't this just memcpy?

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+
+/*
+ * Trace hook for the error_report_start event. If two reports overlap (the
+ * second one report is started before the first one is finished, the first
+ * report is discarded.
+ */
+static void probe_report_start(void *ignore, const char *tool_name,
+			       unsigned long id)
+{
+	atomic_set_release(&current_pid, current->pid);
```
melver, line 51: This should be a cmpxchg:
```
/* Try to become exclusive error reporter; any concurrent errors will be discarded. */
atomic_cmpxchg_relaxed(&current_pid, -1, current->pid);
```

But that doesn't quite work, because we might have a nested task --> interrupt case. If the interrupt reports an error it's going to be a mess, because while it won't succeed in the cmpxchg, it'll start recording info.

I think we should guard against that case, but can probably say if we get nested interrupt a'la task -> interrupt --> interrupt/nmi, then all bets are off and we don't care.

```
/* <explaining what I said above> */
static int get_encoded_pid(void) { return (current->pid << 1) | (!!in_task()); }
```

And then use get_encoded_pid() instead of current->pid throughput. That way nested task -> interrupt will work just fine and the interrupt will fail to cmpxchg, and will also not start recording console output.

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/*
+ * Trace hook for the error_report_end event. If a stale event from the
+ * mismatching error_report_start is received, it is ignored.
+ */
+static void probe_report_end(void *ignore, const char *tool_name,
+			     unsigned long id)
+{
+	unsigned long flags;
```
melver, line 61: We can start ordering these in reverse-xmas tree order. I.e. longest to shortest. Here it'd be pid, then line after is flags.

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+	atomic_set_release(&current_pid, -1);
+}
+
+/*
+ * Trace hook for the console event. If a line comes from a task/CPU that did
+ * not send the error_report_start event, that line is ignored.
+ */
+static void probe_console(void *ignore, const char *buf, size_t len)
+{
+	size_t to_copy;
```
melver, line 86: reverse xmas tree

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
...
+{
+	register_trace_console(probe_console, NULL);
+	register_trace_error_report_start(probe_report_start, NULL);
+	register_trace_error_report_end(probe_report_end, NULL);
+}
+
+static ssize_t last_report_show(struct kobject *kobj,
+				struct kobj_attribute *attr, char *buf)
+{
+	ssize_t ret;
```
melver, line 107: reverse xmas tree

melver (changes requested):

lib/error_report_notify.c
```diff
@@ -0,0 +1,157 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/*
+ * Userspace notification interface for debugging tools.
+ *
+ * Provide two sysfs files:
```
melver, line 6: Isn't this a copy-paste from Kconfig help text?

Hmm, maybe the text here should be the more elaborate one, and the Kconfig help text be a summary. At the very least, I would probably remove the 'poll(POLLPRI)' from the Kconfig help text.

Or should it be the other way around?

I'd probably shorten this text here, as it doesn't show up in Documentation nor Kconfig help.

melver (changes requested):

kernel/trace/error_report-traces.c
ramosian-glider: Done.

include/trace/events/error_report.h
ramosian-glider: How about this:
```
Introduce error_report_start and error_report_end tracepoints.
Those can be used in debugging tools like KASAN, KFENCE or UBSAN
to provide extensions to the error reporting mechanisms (e.g. allow
tests hook into error reporting, ease error report collection from
production kernels).
Another benefit would be making use of ftrace for debugging or
benchmarking the tools themselves.
```
?

>I also wouldn't mention KCSAN here, because it raises the expectation that we'll add support in this series, which I think we shouldn't for now (I'll do that at some point, but not sure yet).

Given the new updates regarding KFENCE, shall I mention it? Or maybe we'd better send the error_report series independently and then send a single followup patch enabling traces in KFENCE?

melver: Sounds good, thanks.


I think we can send all as 1 series. Since this will only be included in the next cycle anyway, we can mention KFENCE. I assume this will also be picked up by -mm, so having the patch enabling traces in KFENCE is fine.

include/trace/events/error_report.h
ramosian-glider: Done. Still gonna mention "tool" in the descriptions though.
(compared to "id", "error_detector" looks a bit bulky in the doc comment, but I don't think that's much of a problem)

mm/kfence/report.c
ramosian-glider: Yes, very good point. We also don't have standard separator length across tools anyway.

include/trace/events/error_report.h
ramosian-glider: Done.

mm/kasan/report.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Changed the title to "lib: add error_report_notify to collect debugging tools' reports" and picked the rest of your text.
Thanks!

lib/Kconfig.debug
ramosian-glider: Hm, won't it be surprising to the users that we truncate the report without mentioning it at least somewhere?
Applied the rest of the text suggestions, will also implement report truncation.

melver: Maybe it's worth mentioning, but I think I just didn't like the calling out of "trimmed at PAGE_SIZE" of something that happens extremely rarely. The way it was put sounds like it's a common occurrence, and might be cause for criticism.

Maybe just: "<something about storing reports> (with maximum report length of PAGE_SIZE)".

ramosian-glider:
Ok, sounds reasonable.

lib/error_report_notify.c
ramosian-glider: We need to return the number of characters written, and scnprintf will do that for us :)

lib/Kconfig.debug
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: It's gone now

mm/kfence/core.c
```diff
@@ -559,12 +559,53 @@ static const struct file_operations objects_fops = {
 	.llseek = seq_lseek,
 };
 
+#define CRASH_BUF_SIZE 16
```
ramosian-glider, line 562: Damn. I did `git checkout HEAD~1` to hide this patch, but apparently it didn't work.
Please disregard it.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+/*
+ * read() handler for /sys/kernel/error_report/last_report.
+ * Because the number of reports can change under our feet, check it again
+ * after copying the report, and retry if the numbers mismatch.
+ * */
+static ssize_t last_report_show(struct kobject *kobj,
+				struct kobj_attribute *attr, char *buf)
+{
+	int index = atomic_read_acquire(&num_reports), old_index;
+	char *out_buf = vmalloc(BUF_SIZE);
```
melver, line 150: Shouldn't the read of the vfs file only succeed when this returns? Is there a chance that somebody might see partial data if we have to redo the copy?

I would check how the return buffer is used, and if there is no chance that some partial copy can become visible, get rid of the temporary buffer.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+	ssize_t ret;
+
+	do {
+		old_index = index;
+		memcpy(out_buf, report_buffer[index % 2], BUF_SIZE);
+		// TODO: need a fence here to prevent reordering with memcpy?
+		index = atomic_read_acquire(&num_reports);
+	} while (index != old_index);
+	ret = scnprintf(buf, BUF_SIZE, out_buf);
```
melver, line 159: Check if strscpy() does what you want here. I think it returns the chars copied not including nul-terminator.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+ * the error_report_start/error_report_end tracing events) and the total report
+ * count.
+ *
+ * Userspace apps can call poll(POLLPRI) on those files to get notified about
+ * the new reports without having to watch dmesg in a loop.
+ *
+ * Because reporting code can be called from low-level routines (e.g. locking
+ * primitives or allocator guts), report recording is implemented using a
+ * seqlock-alike lock-free algorithm.
+ *
```
melver, line 19: remove newline

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+ *  - /sys/kernel/error_report/last_report
+ *  - /sys/kernel/error_report/report_count
+ * that contain the last debugging tool report (taken from dmesg, delimited by
+ * the error_report_start/error_report_end tracing events) and the total report
+ * count.
+ *
+ * Userspace apps can call poll(POLLPRI) on those files to get notified about
+ * the new reports without having to watch dmesg in a loop.
+ *
+ * Because reporting code can be called from low-level routines (e.g. locking
```
melver, line 16: This is a low-level implementation detail. I'd move this above the function implementing last_report_show, which already mentions some of this. Mentioning it here makes it stand out and it feels like it's very important.

There are also similar low-level implementation details, like the cmpxchg on current_pid, that are of equal importance. These details can just be described above the respective functions that implement the majority of the algorithm.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+#include <linux/string.h>
+#include <linux/tracepoint.h>
+#include <linux/vmalloc.h>
+#include <linux/workqueue.h>
+#include <trace/events/error_report.h>
+#include <trace/events/printk.h>
+
+static struct kobject *error_report_kobj;
+
+#define BUF_SIZE PAGE_SIZE
```
melver, line 32: I think we can't have more than PAGE_SIZE because of vfs file size limit, right? If so, a brief comment here would be good.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+
+/*
+ * Trace hook for the error_report_start event. In an unlikely case of another
+ * task already printing a report bail out, otherwise save the current pid
+ * together with in_task() return value.
+ */
+static void probe_report_start(void *ignore, const char *tool_name,
+			       unsigned long id)
+{
+	/* Acquire the writer lock. */
```
melver, line 75: Brief comment that any racing probe_report_start() will fail and not record any report.

We should also mention that we need to roll our own lock using the pid, because probe_console() will be called regardless, and we need to know if the current thread successfully acquired the "lock" or not. If we didn't need to know this, we could have just used a normal spinlock and just trylock here instead.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+	if (pid != get_encoded_pid() || truncated)
+		return;
+
+	idx = (atomic_read_acquire(&num_reports) + 1) % 2;
+	to_copy = min(len, BUF_SIZE - current_report_pos);
+	memcpy(report_buffer[idx] + current_report_pos, buf, to_copy);
+	current_report_pos += to_copy;
+	if (len > to_copy) {
+		truncated = true;
+		memcpy(report_buffer[idx] + current_report_pos -
```
melver, line 128: line-break here worsens readability

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+ * the new reports without having to watch dmesg in a loop.
+ *
+ * Because reporting code can be called from low-level routines (e.g. locking
+ * primitives or allocator guts), report recording is implemented using a
+ * seqlock-alike lock-free algorithm.
+ *
+ */
+
+#include <linux/fs.h>
+#include <linux/string.h>
```
melver, line 23: include <linux/atomic.h>

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+#include <trace/events/error_report.h>
+#include <trace/events/printk.h>
+
+static struct kobject *error_report_kobj;
+
+#define BUF_SIZE PAGE_SIZE
+/* Two buffers to store the finished report and the report being recorded. */
+static char report_buffer[2][BUF_SIZE];
+/*
+ * Total report count. Also serves as a sequence count for report_buffer:
```
melver, line 36: "sequence count" -> "latch"

include/trace/events/error_report.h
melver: Maybe do a s/KASAN, KFENCE or UBSAN/KASAN, KFENCE, *SAN, etc./

because I worry that somebody might call out we're not adding UBSAN support in the series. Trying to manage people's expectations and ensure we don't give them too many reasons to complain...

Unless you feel like adding UBSAN support in this series.

melver (changes requested):
Thanks! I think this is looking good, and we can start polishing. :-)

include/trace/events/error_report.h
ramosian-glider: I would just write "KASAN, KFENCE, etc." then, as *SAN already includes KASAN, and "*SAN and KFENCE" is a little too short.

melver: LGTM

lib/error_report_notify.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Agreed.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+static struct delayed_work reporting_done;
+
+static void error_report_notify(struct work_struct *work)
+{
+	sysfs_notify(error_report_kobj, NULL, "last_report");
+	sysfs_notify(error_report_kobj, NULL, "report_count");
+}
+static DECLARE_DELAYED_WORK(reporting_done, error_report_notify);
+
+static int get_encoded_pid(void)
```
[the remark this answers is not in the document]
ramosian-glider: Done.

lib/Kconfig.debug
```diff
@@ -195,6 +195,19 @@ config DEBUG_BUGVERBOSE
...
 	  debugging but costs about 70-100K of memory.
 
+config ERROR_REPORT_NOTIFY
+	bool "Expose memory error reports to the userspace"
+	depends on TRACING
+	help
+	  When enabled, captures error reports from debugging tools (such as
+	  KFENCE or KASAN)  using console tracing, and exposes reports in
+	  /sys/kernel/error_report/: the file last_report contains the last
+	  report (with maximum report length of PAGE_SIZE), and report_count -
```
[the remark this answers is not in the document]
ramosian-glider: Done.

lib/Kconfig.debug
```diff
@@ -195,6 +195,19 @@ config DEBUG_BUGVERBOSE
...
+	  the total report count.
+	  Userspace programs can call poll(POLLPRI) on those files to get
```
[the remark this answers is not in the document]
ramosian-glider: Done.

lib/Kconfig.debug
```diff
@@ -195,6 +195,19 @@ config DEBUG_BUGVERBOSE
 	  of the BUG call as well as the EIP and oops trace.  This aids
 	  debugging but costs about 70-100K of memory.
 
+config ERROR_REPORT_NOTIFY
+	bool "Expose memory error reports to the userspace"
+	depends on TRACING
+	help
+	  When enabled, captures error reports from debugging tools (such as
+	  KFENCE or KASAN)  using console tracing, and exposes reports in
```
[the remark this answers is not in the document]
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Shall I maybe mention this at the top of the file, where current_pid is declared?

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+	idx = (atomic_read_acquire(&num_reports) + 1) % 2;
+	if (current_report_pos == BUF_SIZE)
+		report_buffer[idx][current_report_pos - 1] = 0;
+	else
+		report_buffer[idx][current_report_pos] = 0;
+
+	atomic_inc_return_release(&num_reports);
+	schedule_delayed_work(&reporting_done, 0);
+	/* Release the writer lock */
+	atomic_set(&current_pid, -1);
```
melver, line 105: This needs to be "release", to ensure the write of num_reports happens before.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+/*
+ * Trace hook for the error_report_start event. In an unlikely case of another
+ * task already printing a report bail out, otherwise save the current pid
+ * together with in_task() return value.
+ */
+static void probe_report_start(void *ignore, const char *tool_name,
+			       unsigned long id)
+{
+	/* Acquire the writer lock. */
+	if (atomic_cmpxchg_relaxed(&current_pid, -1, get_encoded_pid()) != -1)
```
melver, line 76: needs to be acquire

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+static ssize_t last_report_show(struct kobject *kobj,
+				struct kobj_attribute *attr, char *buf)
+{
+	int index = atomic_read_acquire(&num_reports), old_index;
+	char *out_buf = vmalloc(BUF_SIZE);
+	ssize_t ret;
+
+	do {
+		old_index = index;
+		memcpy(out_buf, report_buffer[index % 2], BUF_SIZE);
```
melver, line 155: data_race(memcpy(...)) because we're not using the normal seqlock interface (and add something that on a race this will be a data race, but we'll re-read the data anyway)

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+ * */
+static ssize_t last_report_show(struct kobject *kobj,
+				struct kobj_attribute *attr, char *buf)
+{
+	int index = atomic_read_acquire(&num_reports), old_index;
+	char *out_buf = vmalloc(BUF_SIZE);
+	ssize_t ret;
+
+	do {
+		old_index = index;
```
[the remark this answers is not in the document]
melver: Updated the comment above: we were missing an "smp_rmb()" because the memcpy itself must be ordered before the re-check.

Normal seqlocks also add this. And C++11 seqlocks would use a seq_cst read of some sort. But we don't have a fully-ordered op, and doing a 0-increment would feel wrong.

lib/error_report_notify.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: > I would probably remove the 'poll(POLLPRI)' from the Kconfig help text.
There is no place in Documentation for this line, and I think people need to find this out somehow, so maybe keeping it in Kconfig will help?

melver: Keeping it in Kconfig makes sense, as this file is mainly for kernel devs. So, the redundancy could be resolved by removing it  here and leaving the Kconfig as-is?

I leave it to you. If you think both files need to mention this, I don't mind, but worry that if one of these is updated then the other will be out-of-date, or we have to keep changing 2 places to keep things in sync.

lib/error_report_notify.c
ramosian-glider: Probably not.
Done.

lib/error_report_notify.c
ramosian-glider: Indeed. Done.

lib/error_report_notify.c
```diff
@@ -0,0 +1,207 @@
...
+	if (pid != get_encoded_pid())
+		return;
+
+	idx = (atomic_read_acquire(&num_reports) + 1) % 2;
+	if (current_report_pos == BUF_SIZE)
+		report_buffer[idx][current_report_pos - 1] = 0;
+	else
+		report_buffer[idx][current_report_pos] = 0;
+
+	atomic_inc_return_release(&num_reports);
```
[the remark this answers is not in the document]
ramosian-glider: Done.

lib/error_report_notify.c
ramosian-glider: You are right:

```
sysfs allocates a buffer of size (PAGE_SIZE) and passes it to the
method. Sysfs will call the method exactly once for each read or
write. This forces the following behavior on the method
implementations: 

- On read(2), the show() method should fill the entire buffer. 
  Recall that an attribute should only be exporting one value, or an
  array of similar values, so this shouldn't be that expensive. 

  This allows userspace to do partial reads and forward seeks
  arbitrarily over the entire file at will. If userspace seeks back to
  zero or does a pread(2) with an offset of '0' the show() method will
  be called again, rearmed, to fill the buffer.
```

lib/error_report_notify.c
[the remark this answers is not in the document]
ramosian-glider: Done

lib/error_report_notify.c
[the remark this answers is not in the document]
melver: melted -> combined

lib/error_report_notify.c
```diff
@@ -0,0 +1,225 @@
...
+		/*
+		 * If index and old_index mismatch, we might be accessing
+		 * report_buffer concurrently with a writer thread. In that
+		 * case the read data will be discarded.
+		 */
+		ret = data_race(strscpy(buf, report_buffer[index % 2], BUF_SIZE));
+		/*
+		 * Prevent reordering between the memcpy above and the atomic
+		 * read below.
+		 * See https://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf
```
melver, line 172: We could just refer to our own <include/linux/seqlock.h>, as this just mirrors that implementation. Hans's implementation is very C++11 specific, which doesn't quite apply in LKMM context.

ramosian-glider: Done. I've also rewritten the loop to be more idiomatic.
All in all I decided to not use seqcount_latch_t, which basically just provides a wrapper around an atomic counter, but still requires the user to e.g. calculate the buffer index manually.

melver: That's fair. The most optimal implementation is the one reusing num_reports.

seqcount_latch_t takes care of all the memory-ordering bits, which is important during review. But we'll see if people dislike the open-coded seqlock or not.

lib/error_report_notify.c
```diff
@@ -0,0 +1,224 @@
...
+
+	do {
+		/* Pairs with release in probe_report_end(). */
+		index = atomic_read_acquire(&num_reports);
+		/*
+		 * If index and old_index mismatch, we might be accessing
+		 * report_buffer concurrently with a writer thread. In that
+		 * case the read data will be discarded.
+		 */
+		ret = data_race(
```
melver, line 168: line-break hurts readability.

lib/error_report_notify.c
```diff
@@ -0,0 +1,224 @@
...
+ * Provide two sysfs files:
+ *  - /sys/kernel/error_report/last_report
+ *  - /sys/kernel/error_report/report_count
+ * that contain the last debugging tool report (taken from dmesg, delimited by
+ * the error_report_start/error_report_end tracing events) and the total report
+ * count.
+ *
+ * Because reporting code can be called from low-level routines (e.g. locking
+ * primitives or allocator guts), report recording is implemented using a
+ * seqlock-alike lock-free algorithm.
```
melver, line 15: It *is* a seqlock algorithm -- no need for "alike".

(Also see other comment that's still left above)

lib/error_report_notify.c
```diff
@@ -0,0 +1,224 @@
...
+static void probe_report_start(void *ignore, const char *tool_name,
+			       unsigned long id)
+{
+	/*
+	 * Acquire the writer lock. Any racing probe_report_start will not
+	 * record anything. Pairs with the release in probe_report_end().
+	 */
+	if (atomic_cmpxchg_acquire(&current_pid, -1, get_encoded_pid()) != -1)
+		return;
+	current_pos = 0;
```
melver, line 83: Not sure where the comment went, but there was a comment about 'truncated = false' missing somewhere.

melver (changes requested):
Functionally looks fine (except the missing 'truncated = false').

Some non-functional comments are left, which would be good to resolve or acknowledge.

Thank you!

lib/error_report_notify.c
ramosian-glider: Fixed.

lib/error_report_notify.c
ramosian-glider: Fixed.

lib/error_report_notify.c
ramosian-glider: Put everything on a single line (if the reviewers object, we can break after ` = `)

ramosian-glider:
@melver I think we're all set now - please let me know if you have any other comments before I send this for upstream review.

melver (approved):
There are 3 (outdated) unresolved comments left. Check if they're still relevant.

Then feel free to just send. Any minor bits left can be addressed via LKML review.

Thanks!

lib/error_report_notify.c
ramosian-glider: Moved this part of the comment `probe_report_start()`

lib/error_report_notify.c
ramosian-glider: Yes, you are right. Added the comment.

ramosian-glider (commented):
Found only two remaining comments, addressed them.

melver (approved):
LGTM

Assuming you already run tests and subjected it to syzkaller torture testing.

melver:
> LGTM
> 
> Assuming you've already run tests and subjected it to syzkaller torture testing.

Ah, to torture test the code here we need lots of bugs. 

   CONFIG_KFENCE_STRESS_TEST_FAULTS=<non-zero> 

would probably do the job.

lib/error_report_notify.c
```diff
@@ -0,0 +1,259 @@
...
+ * Skip one or two leading pair of brackets containing the log timestamp and
+ * the task/CPU ID from the report line, e.g.:
+ *   [    0.698431][    T7] BUG: KFENCE: use-after-free ...
+ * becomes:
+ *   BUG: KFENCE: use-after-free ...
+ *
+ * Report size is only 4K, and this boilerplate can easily account for half of
+ * that amount.
+ */
+const char *skip_meta(const char *buf, size_t len)
```
melver, line 131: static

lib/error_report_notify.c
```diff
@@ -0,0 +1,259 @@
...
+ * Trace hook for the console event. If a line comes from a task/CPU that did
+ * not send the error_report_start event, that line is ignored. Otherwise, it
+ * is stored in the report_buffer[(num_reports + 1) % 2].
+ */
+static void probe_console(void *ignore, const char *buf, size_t len)
+{
+	int pid = atomic_read(&current_pid);
+	size_t to_copy;
+	int idx;
+	const char *skip;
```
melver, line 160: "skip" sounds like it should be a count. If this is the string without the meta information, it probably should be something like "without_meta".


An alternative: make "skip_meta()" actually do something, which is what I expected it to do given its current name:
```
...
skip_line_meta(&buf, &len); // mutates both buf and len
...
```

Whichever you think is cleaner. I suggest at least changing skip_meta() name to something more descriptive, because "meta" could be anything.

lib/error_report_notify.c
```diff
@@ -0,0 +1,259 @@
...
+ *   [    0.698431][    T7] BUG: KFENCE: use-after-free ...
+ * becomes:
+ *   BUG: KFENCE: use-after-free ...
+ *
+ * Report size is only 4K, and this boilerplate can easily account for half of
+ * that amount.
+ */
+const char *skip_meta(const char *buf, size_t len)
+{
+	const char *found;
```
melver, line 133: reverse xmas tree

lib/error_report_notify.c
```diff
@@ -0,0 +1,259 @@
...
+	int i;
+	int num_brackets = IS_ENABLED(CONFIG_PRINTK_TIME) +
+			   IS_ENABLED(CONFIG_PRINTK_CALLER);
+
+	for (i = 0; i < num_brackets; i++) {
+		if (!len || (buf[0] != '['))
```
melver, line 139: Verify that KERN_ERR, KERN_WARN etc don't appear in the "buf" we receive. Because if they do, they likely are at buf[0].

melver, line 139: Braces around 'buf[0] != ...' not necessary.

melver (changes requested):

lib/error_report_notify.c
```diff
@@ -0,0 +1,259 @@
...
+ * that amount.
+ */
+static const char *skip_meta(const char *buf, size_t len)
+{
+	const char *found;
+	int i;
+	int num_brackets = IS_ENABLED(CONFIG_PRINTK_TIME) +
+			   IS_ENABLED(CONFIG_PRINTK_CALLER);
+
+	for (i = 0; i < num_brackets; i++) {
```
melver, line 138: Hmm, it seems 'i' is redundant. Could just be 'while (num_brackets--)' ?

lib/error_report_notify.c
ramosian-glider: I took the approach that mutates buf/len, thanks!

lib/error_report_notify.c
ramosian-glider: Fixed.

lib/error_report_notify.c
ramosian-glider: Looks like the trace already receives the line without the log level.

ramosian-glider: Fixed.

lib/error_report_notify.c
ramosian-glider: Fixed.

ramosian-glider:
I've also noticed that a buffer captured by console tracing may contain \n's, in which case we need to handle every line separately. Fixed that as well.

lib/error_report_notify.c
```diff
@@ -0,0 +1,278 @@
...
+		memcpy(report_buffer[idx] + current_pos, buf, to_copy);
+		current_pos += to_copy;
+		if (cur_len > to_copy) {
+			truncated = true;
+			memcpy(report_buffer[idx] + current_pos - sizeof(TRUNC_MSG),
+			       TRUNC_MSG, sizeof(TRUNC_MSG));
+			break;
+		}
+		buf += cur_len;
+	} while (len);
```
melver, line 194: Why not 

```
while (len) {
   ...
}
```

Because if len==0, we shouldn't be doing anything anyway.

melver (approved):
LGTM. Please send if it all works now and has been stress tested.

lib/error_report_notify.c
ramosian-glider: Fixed. I also noticed my code was incorrectly deleting the leading space - fixed that as well.

PR #173, opened 2020-12-21
Joern's suggestions

melver:
These had already been posted in on kasan-dev in the discussion with Joern. I'll send them after 5.11-rc1.

### DIFF
--- a/mm/kfence/core.c
+++ b/mm/kfence/core.c
@@ -88,14 +88,16 @@ struct kfence_metadata kfence_metadata[CONFIG_KFENCE_NUM_OBJECTS];
 static struct list_head kfence_freelist = LIST_HEAD_INIT(kfence_freelist);
 static DEFINE_RAW_SPINLOCK(kfence_freelist_lock); /* Lock protecting freelist. */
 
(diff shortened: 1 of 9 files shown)


mm/kfence/core.c
```diff
@@ -88,14 +88,16 @@ struct kfence_metadata kfence_metadata[CONFIG_KFENCE_NUM_OBJECTS];
...
 
-/* Gates the allocation, ensuring only one succeeds in a given period. */
-static atomic_t allocation_gate = ATOMIC_INIT(1);
-
 /* Wait queue to wake up allocation-gate timer task. */
 static DECLARE_WAIT_QUEUE_HEAD(allocation_wait);
+#endif
+
+/* Gates the allocation, ensuring only one succeeds in a given period. */
+atomic_t kfence_allocation_gate = ATOMIC_INIT(1);
```
ramosian-glider, line 100: We are not exporting the allocation gate, any reason to add the `kfence_` prefix?

ramosian-glider:
LGTM otherwise.

mm/kfence/core.c
melver: It's no longer static and may now clash with any other global.

Issue #172, opened 2020-12-17
It's unclear how to activate kasan


exx8:
in the doc it's written shortly, that to activate kasan one should:
`CONFIG_KASAN=y`
but it's unclear how and where to do it.
I hope you can help and clarifiy.
thanks

PR #174, opened 2020-12-23
Link in readme to an install guide 

exx8:
I know that for the developer of KASAN it's clear that you need to build the kernel for KASAN to work. But I believe a full guide can be helpful, for those who are new to KASAN.

This PR adds a link to  a full guide that clarify how to build kernel with KASAN.
solves #172

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 KernelAddressSanitizer (KASAN) is a dynamic memory error detector for the Linux kernel that allows to find use-after-free and out-of-bounds bugs.
 
-KASAN is available in the mainline Linux kernel starting from version `4.0` and can be enabled with `CONFIG_KASAN=y`.
+KASAN is available in the mainline Linux kernel starting from version `4.0` and can be enabled with `CONFIG_KASAN=y`, for more details please read [this article](https://www.programmersought.com/article/7010851656/)
 
 See more details in [KASAN documentation](https://www.kernel.org/doc/html/latest/dev-tools/kasan.html).
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 KernelAddressSanitizer (KASAN) is a dynamic memory error detector for the Linux kernel that allows to find use-after-free and out-of-bounds bugs.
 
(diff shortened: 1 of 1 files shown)


exx8:
@googlebot I signed it!

ramosian-glider:
I think it is pretty clear from https://www.kernel.org/doc/html/latest/dev-tools/kasan.html (referenced just below the part you are changing) that one must build the kernel in order for CONFIG_KASAN to take effect.
There are many online instructions on how to build the kernel, and someone wanting to fix bugs in it will be probably able to find one.

exx8:
> I think it is pretty clear from https://www.kernel.org/doc/html/latest/dev-tools/kasan.html (referenced just below the part you are changing) that one must build the kernel in order for CONFIG_KASAN to take effect.
> There are many online instructions on how to build the kernel, and someone wanting to fix bugs in it will be probably able to find one.

see:
https://unix.stackexchange.com/questions/580865/how-do-i-enable-a-configuration-in-a-kernel

KASAN is a must for a Linux kernel module programmer. Because, it's a bit exotic, probably there are large number of one time kernel module writers.
Currently, it is not clear from the instructions that you need to rebuild the kernel with this option, nor how to do this. This PR tries, to make the first time, a bit easier.

dvyukov:
KASAN is now upstream and docs are maintained upstream. The README page specifically only provides a few key references. If there is something missing in the upstream doc or they can be improved, I think the improvements should go there:
https://www.kernel.org/doc/html/latest/dev-tools/kasan.html
We don't want any active development/maintenance on the github page. Perhaps we need to reduce the docs even more to make it clearer that this is just a stub?

xairy:
> KASAN is now upstream and docs are maintained upstream. The README page specifically only provides a few key references. If there is something missing in the upstream doc or they can be improved, I think the improvements should go there:
> https://www.kernel.org/doc/html/latest/dev-tools/kasan.html
> We don't want any active development/maintenance on the github page. Perhaps we need to reduce the docs even more to make it clearer that this is just a stub?

Done: https://github.com/google/kasan/commit/c0f2681e7d9371caede2458dbb9fe6fa7ac8a0ab

PR #176, opened 2021-01-16
kfence: add missing copyright and description headers

melver:


ramosian-glider:
You beat me to it :)
We'll probably need to split this patch though, as it is touching multiple upstream-to-be patches.
LGTM anyway.

